PR #50, opened 2014-02-05
fix issue 48

AntonioGonzalez:
Added two protected methods to DefaultClusterRenderer to get a marker from ClusterItem and vice versa


broady:
Obsolete, right? Re-open if not.

PR #55, opened 2014-02-10
Made minimum cluster size configurable

choefele:
Please consider merging this as it will make `DefaultClusterRenderer` more flexibel.


stephenmcd:
Thanks @choefele

PR #81, opened 2014-04-23
Zoom in the cluster

isatimur:
The code in comments may help people easy implement clusters zoom in.
So, please, check is it ok for you team or not?


stephenmcd:
Thanks @isatimur

PR #155, opened 2015-02-13
KML Importing Utility

lchans:


markmcd:
Merged manually in d606fcd.

PR #156, opened 2015-02-13
GeoJSON Importing Utility

juliawong:
:tada:


markmcd:
Merged manually in 46a4e20.

Issue #133, opened 2014-11-20
Ability to remove Marker from NonHierarchicalDistanceBasedAlgorithm


bedzinsa:
Hello,

I am using NonHierarchicalDistanceBasedAlgorithm and GridBasedAlgorithm is not suitable for me. However i really need the removeItem Method from GridBasedAlgorithm which in NonHierarchical throws an Exception. I tried to extend the hierarchical algorithm and override the removeItem method and remove item using the same principle as it is displayed in addItem method.

    @Override
    public void removeItem(MarkerItem item)
    {
        final QuadItem<MarkerItem> quadItem = new QuadItem<MarkerItem>(item);

        synchronized (mQuadTree)
        {
            mItems.remove(quadItem);
            mQuadTree.remove(quadItem);
        }
    }

However, im not sure whether it is remove or not that way, because when i call removeItem, and recluster items it is still displayed on the map. But if i try it twice, it throws me nullpointerexception on
final QuadItem<MarkerItem> quadItem = new QuadItem<MarkerItem>(item); which I think means that it no longer exists. 

Any idea how should i do it properly?

stephenmcd:
f49f604ee1262105a40ca61bcae90f217f22043c

PR #169, opened 2015-06-04
Equals for Cluster subclasses

hypest:
Adds implementations of "equals()" and "hashCode()" in StaticCluster and QuadItem classes. Benefit is the ability to reliably compare (and identify) instances of those classes.

As a bonus (and an implicit way to test QuadItem::hashCode() without exposing the inner class) this PR also includes an implementation of NonHierarchicalDistanceBasedAlgorithm::removeItem(). By the way, this would effectively fix issue #133.

PR commits include relevant tests. Let me know if more testing should be added.

Also, the added test files are new and thus copyrighted after myself, but still under the Apache v2 license. Let me know if this needs to be revised somehow.

Cheers!

### DIFF
--- a/library/src/com/google/maps/android/clustering/algo/NonHierarchicalDistanceBasedAlgorithm.java
+++ b/library/src/com/google/maps/android/clustering/algo/NonHierarchicalDistanceBasedAlgorithm.java
@@ -183,5 +183,19 @@ public class NonHierarchicalDistanceBasedAlgorithm<T extends ClusterItem> implem
         public int getSize() {
             return 1;
         }
+
+        @Override
+        public int hashCode() {
+            return mClusterItem.hashCode();
+        };
+
+        @Override
+        public boolean equals(Object other) {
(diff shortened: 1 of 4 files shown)


hypest:
I signed it!

markmcd:
Code looks good - thanks for taking the time to write tests, too.

With the copyright headers, can you please change the (c) line to `Copyright 2015 Google Inc.`? You are more than welcome to add yourself to the CONTRIBUTORS.md file (but you don't have to if you don't want to).

hypest:
Thanks for reviewing @markmcd!

More than happy to make the changes! Thing is, I no longer have commit access to this fork so I can't commit the changes. For that, I opened a new PR (https://github.com/googlemaps/android-maps-utils/pull/245).

Let me know if there's anything more I should do with the current PR.

Thanks in advance!

markmcd:
Dropping in favour of #245.

Issue #142, opened 2015-01-08
libpng warning: iCCP: Not recognizing known sRGB profile that has been edited


johnjohndoe:
Gradle outputs the above *libpng* warning for the following files when *android-maps-utils
 v.0.3.4* is used in an Android application project:

```
com.google.maps.android/android-maps-utils/0.3.4/res/
    drawable-xhdpi-v4/bubble_shadow.9.png
    drawable-xxhdpi-v4/bubble_mask.9.png
    drawable-xhdpi-v4/bubble_mask.9.png
    drawable-xxhdpi-v4/bubble_shadow.9.png
```

This warning is already discussed in [issue #77704](https://code.google.com/p/android/issues/detail?id=77704).

broady:
Huh, I did notice that. What's your recommended fix? Should we strip the color profile from the images?

johnjohndoe:
Not sure. There are a bunch of discussion about the topic in the net (also apart from Android) if you search for it. I could not find a consent, though.

PR #176, opened 2015-06-17
Remove sRGB color profile from PNGs

wesleym:
Fixes googlemaps/android-maps-utils#142 as described in
https://code.google.com/p/android/issues/detail?id=77704#c25


wesleym:
Typo in commit message.

PR #177, opened 2015-06-17
Remove sRGB color profile from PNGs

wesleym:
Fixes googlemaps/android-maps-utils#142 using the solution described in
https://code.google.com/p/android/issues/detail?id=77704#c25


wesleym:
I signed it! (as the company named "Nextdoor")

PR #199, opened 2015-08-27
Make getColor protected in DefaultClusterRenderer

jyaganeh:
I'd like to use the `DefaultClusterRenderer` in my project but override the colors.

This PR changes the access level of the `getColor` method so that it can be overridden in subclasses of `DefaultClusterRenderer`.


jyaganeh:
I signed it!

PR #144, opened 2015-01-10
Add Douglas-Peucker algorithm

pvllnspk:

PR #201, opened 2015-09-16
Add Douglas-Peucker line simplification algorithm as PolyUtil.simplify()

barbeau:
* Douglas-Peucker implementation ported from the Google MyTracks project (https://code.google.com/p/mytracks/source/browse/MyTracks/src/com/google/android/apps/mytracks/util/LocationUtils.java#81), (c) Google 2008, licensed under Apache v2.0.
* Add demo activity for `PolyUtil.simplify()` - shows the original line in black, as well as simplified versions of the line with multiple tolerances in different colors
* Add unit tests for `PolyUtil.simplify()` and `PolyUtil.distanceToLine()` in `PolyUtilTest`

As discussed with @broady in https://github.com/googlemaps/android-maps-utils/pull/144#issuecomment-140505017.

@broady A few unrelated white space changes snuck into this commit via Android Studio - seems this happens on save, and I haven't found a way to disable it.  Let me know if this is an issue - I'm using the official Android code template.

### DIFF
--- a/library/src/com/google/maps/android/PolyUtil.java
+++ b/library/src/com/google/maps/android/PolyUtil.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013 Google Inc.
+ * Copyright 2008, 2013 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,7 +20,9 @@ import com.google.android.gms.maps.model.LatLng;
 
 import java.util.List;
 import java.util.ArrayList;
+import java.util.Stack;
 
+import static com.google.maps.android.SphericalUtil.*;
 import static java.lang.Math.*;
 import static com.google.maps.android.MathUtil.*;
 
@@ -38,11 +40,11 @@ public class PolyUtil {
 
     /**
      * Returns mercator(latitude-at-lng3) on the Rhumb line (lat1, lng1) to (lat2, lng2). lng1==0.
-     */        
+     */
     private static double mercatorLatRhumb(double lat1, double lat2, double lng2, double lng3) {
         return (mercator(lat1) * (lng2 - lng3) + mercator(lat2) * lng3) / lng2;
     }
-   
+
     /**
      * Computes whether the vertical segment (lat3, lng3) to South Pole intersects the segment
      * (lat1, lng1) to (lat2, lng2).
@@ -84,7 +86,7 @@ public class PolyUtil {
             tan(lat3) >= tanLatGC(lat1, lat2, lng2, lng3) :
             mercator(lat3) >= mercatorLatRhumb(lat1, lat2, lng2, lng3);
     }
-    
+
     /**
      * Computes whether the given point lies inside the specified polygon.
      * The polygon is always cosidered closed, regardless of whether the last point equals
@@ -201,7 +203,7 @@ public class PolyUtil {
             for (LatLng point2 : poly) {
                 double lat2 = toRadians(point2.latitude);
                 double y2 = mercator(lat2);
-                double lng2 = toRadians(point2.longitude);                
+                double lng2 = toRadians(point2.longitude);
                 if (max(lat1, lat2) >= minAcceptable && min(lat1, lat2) <= maxAcceptable) {
                     // We offset longitudes by -lng1; the implicit x1 is 0.
                     double x2 = wrap(lng2 - lng1, -PI, PI);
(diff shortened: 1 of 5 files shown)


broady:
Whitespace changes LGTM. Rest LGTM, too.

pinging @googlebot for CLA ... (I don't think this actually does anything...)

@jfschmakeit @markmcd any thoughts?

broady:
CLA checked manually. :+1:

library/src/com/google/maps/android/PolyUtil.java
```diff
@@ -284,6 +286,102 @@ private static boolean isOnSegmentGC(double lat1, double lng1, double lat2, doub
...
     /**
+     * Simplifies the given line using the Douglas-Peucker decimation algorithm.  Increasing the
+     * tolerance will result in fewer points in the simplified line.
+     *
+     * @param line input line to be simplified
+     * @param simplifiedLine a simplified line produced by the Douglas-Peucker algorithm
+     * @param tolerance in meters.  Increasing the tolerance will result in fewer points in the
+     *                  simplified line.
+     */
+    public static void simplify(List<LatLng> line, List<LatLng> simplifiedLine, double tolerance) {
```
markmcd: Why are you using out-parameters (and void return type), instead of returning a LatLng array or List?

barbeau: I kept the same method signature and return type from the original MyTracks implementation.  However, I'd actually prefer returning a `LatLng` List, so if you agree I'd be happy to change it.

broady: +1, that's preferable. Also, change the parameter name to `polyline` to be consistent with the other functions in this file.

broady: Actually - does this work for closed loops (i.e. polygon simplification)?

barbeau: That's a good question - I think so, although I'll run some tests to see for sure.

broady: Looks like it doesn't modify the lat/lons themselves, so we just need to be sure that the first and last points are unmodified.

I can't tell just by reading the code.

barbeau: Yes, `assertEndPoints()` test checks that the first and last points are unmodified.

barbeau:
@broady @markmcd Ok, just force-pushed a new commit into this PR to replace the old one.  This changes to the method signature and return type to:
`public static List<LatLng> simplify(List<LatLng> polyline, double tolerance) {`

Don't know if I'll get to testing if this works for polygons tonight yet.

markmcd:
LGTM. Thanks for this - it's a valuable contribution.

Let us know when you're done so we can merge.

barbeau:
Looks like this will work for polygons too, but there are a few caveats:

1.  **The polygon we run the algorithm on can't be closed (i.e., first and last `LatLng` in the List can't be exactly the same)** - this breaks Douglas-Peuker, as it requires a positive distance measurement to start the algorithm.
2.  **If we remove the last coordinate duplicate coordinate to open the polyline, it doesn't fully simplify the polygon (there is one extra point)** - you can see this in the screenshots below with a simplified triangle.  The polyline starts with the lower-left corner - blue is the original triangle, and yellow is the simplified version.  In the left screenshot, the bottom edge of the triangle still has a point in the middle (the last point prior to closing the polygon) that should have been removed based on the tolerance.  The right screenshot shows the correctly simplified yellow triangle, with the middle point in the bottom edge removed.

image

I produced the yellow polygon on the right by adding a very small offset to the final "duplicate" point that closes the polygon.  So, it looks like:

~~~
ArrayList<LatLng> triangle = new ArrayList<>();
triangle.add(new LatLng(28.06025,-82.41030));  // First point
...
triangle.add(new LatLng(28.06025000001, -82.41030000001));  // Closing "duplicate" point
~~~

This seems to work without causing problems - here's the same result for an oval:

image

So, as far as handling polygons with the `simplify()` method, I think we have two options:

1. **Require developers to pass in a closed polygon to the `simplify()` method** - Inside the method, we'd check to see if the first and last point were the same, and if so, add the offset to the last point.  We'd add the requirement to the Javadocs for passing in a closed polygon.
2. **Add a new `simplifyPolygon()` method** - We'd allow developers to pass in closed or open polygons into this method.  If its closed, we'd add the offset to the last point, and if its open, we'd close it with the offset point - then we'd call `simplify()`.

For the output of either option above, we would swap out the final point w/ the offset for the original final point before returning the list.

Thoughts?  I think the first option is my preference, as it keeps the library simpler - worst case scenario, a developer passes in an unclosed polygon, and they get a polygon back that's not totally simplified.  This won't even be noticeable in most cases, as it's only a single extra point.

barbeau:
Also, if we handle both polylines and polygons in the same `simplify()` method, I propose we change the parameter name to `shape`:
`public static List<LatLng> simplify(List<LatLng> shape, double tolerance)`

barbeau:
@broady @markmcd I went ahead and implemented the Option 1 above ("Require developers to pass in a closed polygon to the `simplify()` method"), and force-pushed a new commit for this into this PR to replace the old ones.  So, simplification for polylines and polygons are now both fully supported.  I also included a number of new unit tests to cover polygon simplification as well.

So, it should be ready for your final review, as long as you're good with the above design decisions.

library/src/com/google/maps/android/PolyUtil.java
```diff
@@ -284,6 +286,150 @@ private static boolean isOnSegmentGC(double lat1, double lng1, double lat2, doub
...
+     * The time complexity of Douglas-Peucker is O(n^2), so take care that you do not call this
+     * algorithm too frequently in your code.
+     *
+     * @param shape shape (polyline or polygon) to be simplified.  Polygon should be closed (i.e.,
+     *              first and last points should have the same latitude and longitude).
+     * @param tolerance in meters.  Increasing the tolerance will result in fewer points in the
+     *                  simplified shape.
+     * @return a simplified shape produced by the Douglas-Peucker algorithm
+     */
+    public static List<LatLng> simplify(List<LatLng> shape, double tolerance) {
```
broady: s/shape/poly/

broady:
Thank you for the diligent work on this.

LGTM. Just one small nit: I think `shape` should be renamed to `poly`.

barbeau:
@broady Ok, np, I'll push a new commit shortly.

barbeau:
Ok, done.  I changed a few other `shape` references to `poly` as well.

barbeau:
Awesome, thanks @broady and @markmcd for the review/merge!

Would it be possible to get a new library release out with this included?  New demo app release would be great too.

broady:
yup - I'm pretty busy with a bunch of other stuff, so I could only do a release next week. Maybe @markmcd will get to it first. We'll coordinate.

barbeau:
:+1:

barbeau:
Just to make sure this doesn't fall through the cracks - before this is released, PR https://github.com/googlemaps/android-maps-utils/pull/205 should be reviewed/merged.  That PR fixes an issue with polygon simplification.

barbeau:
Friendly ping - @broady and @markmcd do you think you could get a new release out soon?

On an unrelated note - I'm prepping a release of a major app overhaul (including port to Android Maps v2), but seeing a bizarre issue with markers jumping around on the map that's bad enough to postpone release - here's a video:
https://youtu.be/gYBunJYZTMo

So far no traction with an [AOSP issue](https://code.google.com/p/gmaps-api-issues/issues/detail?id=8455) I filed in August.  Any ideas for next steps?

barbeau:
Another friendly ping -  @broady and @markmcd could we get a new release that includes PolyUtil.simplify()?

barbeau:
Yet another friendly ping - @broady and @markmcd could we get a new release that includes PolyUtil.simplify()?

markmcd:
Sure thing. I was hoping to have contributed a little more myself to the next release, but I've pushed 0.4.1 anyways.

barbeau:
Awesome, thanks!!

PR #203, opened 2015-09-18
Added travis test support

markmcd:
May not work yet - making sure the config file works first.

PR #209, opened 2015-10-05
Fix #207 - Add Travis CI support

### LINKED ISSUE #207: Add Travis CI builds for PRs?
IMHO adding [Travis CI](https://travis-ci.org/) support to this project would be helpful - it would help reveal issues like #187 immediately, rather than having to track down their cause later.

@broady @markmcd Thoughts?

If you're willing to turn on Travis support for this project in the Travis CI console, I'd be willing to try configure the `.travis.yml` script.  I've done this on many Android apps, but never on a library before, and its always good to learn new things :).

barbeau:
~~As discussed in https://github.com/googlemaps/android-maps-utils/issues/207#issuecomment-144904766, requires changing to `compileSdkVersion "Google Inc.:Google APIs:21"`, because:
a) Google API emulator image is needed to run tests that require Google Play Services (see https://developers.google.com/android/guides/setup)
b) the Google API Level 23 emulator seems to have a bug that doesn't allow installing/running tests from the command line~~

~~This patch also adds a command-line option to disable pre-dexing, which is used on Travis to decrease build times (see http://tools.android.com/tech-docs/new-build-system/tips#TOC-Improving-Build-Server-performance.)~~

See below comment for revised PR contents, based on comments on the initial commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,47 @@
+language: android
+jdk: oraclejdk7
+# Use the Travis Container-Based Infrastructure
+sudo: false
+
+cache:
+  directories:
+    - ${TRAVIS_BUILD_DIR}/gradle/caches/
+    - ${TRAVIS_BUILD_DIR}/gradle/wrapper/dists/
+
+env:
+  global:
+    - ANDROID_API_LEVEL=23
+    - EMULATOR_API_LEVEL=21
+    - ANDROID_BUILD_TOOLS_VERSION=23.0.0
+    - ANDROID_ABI=google_apis/armeabi-v7a
+    - ADB_INSTALL_TIMEOUT=20 # minutes (2 minutes by default)
+
+android:
+  components:
+    - platform-tools
+    - tools
+    - build-tools-$ANDROID_BUILD_TOOLS_VERSION
+    - android-$ANDROID_API_LEVEL
+    - android-$EMULATOR_API_LEVEL
+    # For Google APIs
+    - addon-google_apis-google-$ANDROID_API_LEVEL
+    - addon-google_apis-google-$EMULATOR_API_LEVEL
+    # Google Play Services
+    - extra-google-google_play_services
+    # Support library
+    - extra-android-support
+    # Latest artifacts in local repository
+    - extra-google-m2repository
+    - extra-android-m2repository
+    # Specify at least one system image
+    - sys-img-armeabi-v7a-addon-google_apis-google-$ANDROID_API_LEVEL
+    - sys-img-armeabi-v7a-addon-google_apis-google-$EMULATOR_API_LEVEL
+
+before_script:
+  # Create and start emulator
+  - echo no | android create avd --force -n test -t "Google Inc.:Google APIs:"$EMULATOR_API_LEVEL --abi $ANDROID_ABI
+  - emulator -avd test -no-skin -no-audio -no-window &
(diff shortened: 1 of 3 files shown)


demo/build.gradle
```diff
@@ -18,7 +18,7 @@ dependencies {
 }
 
 android {
-    compileSdkVersion 23
+    compileSdkVersion "Google Inc.:Google APIs:21"
```
markmcd: Hmm I'd prefer to keep this on the latest API level. Can you set it to '23' here and override these when it's run through Travis?

barbeau: I assume that's possible, although I've never configured that setup before.  Are you able to see if anyone on the Android team can explain why the API 23 Google API emulator isn't working with gradle commands?

http://stackoverflow.com/questions/32952413/gradle-commands-fail-on-api-23-google-api-emulator-image-armeabi-v7a

If the emulator issue isn't something that's going to be fixed anytime soon I can look into running on API 21 only on Travis.

barbeau: Actually, I figured out a way to get this to run on API 21 emulator via the travis script itself.  I'll force-push a new commit into this PR that keeps the project `compileSdkVersion 23` while running on the API Level 21 Google API emulator.

barbeau:
@markmcd Ok, I just force-pushed a new commit into this PR to replace the old one.  Here is the new PR/commit description:

* This patch keeps the project at `compileSdkVersion 23` while running on the API Level 21 Google API emulator on Travis, because:

 a) Google API emulator image is needed to run tests that require Google Play Services (see https://developers.google.com/android/guides/setup, https://github.com/googlemaps/android-maps-utils/pull/203#issuecomment-141721720)

 b) API Level 23 Google API emulator seems to have a bug preventing use on Travis - see https://github.com/googlemaps/android-maps-utils/issues/207#issuecomment-144904766
* When the issue with the API Level 23 Google API emulator is resolved, the `EMULATOR_API_LEVEL` variable in .travis.yml can be bumped to 23 as well.
* This patch also adds a command-line option to disable pre-dexing, which is used on Travis to decrease build times (see http://tools.android.com/tech-docs/new-build-system/tips#TOC-Improving-Build-Server-performance.)

This is a much better solution, as it doesn't alter the project build config at all, but still allows it to run on Travis.

barbeau:
Well, hold that thought...For some reason this is failing now, even though what should be the same config [passed earlier](https://travis-ci.org/barbeau/android-maps-utils/builds/83955408).  Will update again shortly...

barbeau:
I think this should be ok to merge.  Seems Travis had an existential crisis in the middle of the [initial build of this branch](https://travis-ci.org/barbeau/android-maps-utils/builds/83980263).  It initially failed, and when I tried to restart, it seems like its hanging indefinitely in the "created" state but refuses to start.  I created a [new branch](https://github.com/barbeau/android-maps-utils/tree/travis-rebase-test) from the [exact same commit](https://github.com/barbeau/android-maps-utils/commit/807f515f535a4cea06ed8b4299548ed74350a126), and this [passes on Travis](https://travis-ci.org/barbeau/android-maps-utils/builds/83987420).

So, this should be ready for review/merge.

markmcd:
LGTM

markmcd:
Oh - and thanks again for doing this! It's huge :+1:

barbeau:
No prob! :) Thanks for the review/merge.

PR #210, opened 2015-10-06
Add containsLocation method to KmlGeometry and Placemark classes.

homaralex:
Convenience methods for checking whether a given location is inside a geometry or placemark object,
using the PolyUtil.containsLocation method.

### DIFF
--- a/library/src/com/google/maps/android/kml/KmlContainsLocation.java
+++ b/library/src/com/google/maps/android/kml/KmlContainsLocation.java
@@ -1,0 +1,12 @@
+package com.google.maps.android.kml;
+
+
+import com.google.android.gms.maps.model.LatLng;
+
+public interface KmlContainsLocation {
+    /**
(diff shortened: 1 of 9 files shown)


homaralex:
I signed it!

broady:
This doesn't really make sense for lines or points. What is the motivation behind adding this?

homaralex:
I needed a way to check inside, given a location, in which areas of a KML map does it lie - and then get Placemark info etc.
Maybe I was just missing a simpler way?

broady:
That kind of code probably belongs *outside* these classes, I think. I can see something like `bool intersects(LatLngBounds)` being something generally applicable, though.

homaralex:
The `LatLngBounds` approach wasn't accurate enough for me.
But yes, maybe this could be in `PolyUtil` or somewhere in KML Containers.

Issue #82, opened 2014-05-02
Is there a way to just renderer the visible clusters and items?


coyarzun89:
Hi,
I don't know if it is the default behavior, but both methods, onBeforeClusterItemRendered and onBeforeClusterRendered, are called every time that I zoom out the map, even when there are not clusters or items visible. Is there a way to just renderer the visible clusters and items?
I have a map with 1000 items, and every item has his own image, so I need to render only the visible elements.
Thanks!

------
From duplicate issue #82, from @kaads123:

**Summary:** 

ClusterManager draws all the markers/clusters, not only visible ones. It should draw only those that are in visibleRegion. This causes a map with 2000 markers skip frames and ANR (even on new devices) when zooming in/out.

**Steps to reproduce:** 

Add 2000 or more markers and try zooming in and zooming out fast. This can be reproduced in the sample app

**Expected behavior:** 

Draw only those markers/clusters which are in visibleRegion.latLngBounds of the GoogleMap

**Observed behavior:** 

Draws every marker added to ClusterManager

**Device and Android version:** 

Moto G3 with Android 6.0.1
Nexus 6P with Android 7.1.2

**Screenshots:** 

![screenshot_20170628-182053](https://user-images.githubusercontent.com/13378420/27645450-c47d1f4e-5c2e-11e7-9b33-8a9e2907e6b8.png)

Zoomed out from 20f to 1f quickly

multivoltage:
@zamesilyasa ,I found your code. Sorry....but how I can edit Clustermanager?

zamesilyasa:
I honestly don't understand what are you trying to do

multivoltage:
I found your repo, which is fork of original google-maps-utilis. I cloned issue-82 branch in my pc. With Android Studio I add that project as module. Then i remove original gradle dependency for google-maps-utils. At this point I can use all classs like before. But i don't understand Clustmanager. I was thinking to find `.setClusterOnlyVisibleArea(true);` method. I noticed that you ClusterManager class is different from that in this link:
[https://github.com/zamesilyasa/android-maps-utils/commit/1cf9f4e553fded253b0fcb8d31f58640687de985](url)

So, maybe I jump some step, I don't know

zamesilyasa:
That's because there is no such method anymore. Please, check VisibleClusteringDemoActivity.
The only thing you have to do, to use my patch is to set NonHierarchicalViewBasedAlgorithm to your cluster manager instance, that's it.

[...]

PR #217, opened 2015-10-19
Render only visible items on the map

zamesilyasa:
Implemented rendering only visible items on the map(issue #82 )
Added VisibleClusteringDemoActivity activity which demonstrates clustering of 20k items.

[Review on Reviewable](https://reviewable.io/reviews/googlemaps/android-maps-utils/217)


barbeau:
@broady or @markmcd could you rerun [this build on Travis](https://travis-ci.org/googlemaps/android-maps-utils/builds/86192265)?  The initial build failed with `com.android.ddmlib.ShellCommandUnresponsiveException`, which isn't related to the PR contents.

barbeau:
Thanks, looks like the Travis build passed now.  I've been seeing occasional failures due to `com.android.ddmlib.ShellCommandUnresponsiveException` on Travis builds on other projects too.  Increasing the timeout doesn't seem to help, so right now unfortunately the workaround is just to poke Travis again, and it usually passes on the second attempt.

meierjan:
+1 for merge

Dima564:
+1

broady:
Mark or @jfschmakeit can you look at this more closely?

markmcd:
Awesome - thanks for this contribution. It's looking great on the device. Just a couple of comments around the code.

Oh and I'll re-run the Travis builds when needed, they're very flaky at the moment.

library/src/com/google/maps/android/clustering/ClusterManager.java
```diff
@@ -92,13 +93,23 @@ public void setAlgorithm(Algorithm<T> algorithm) {
...
         }
+
+        if (mAlgorithm instanceof GoogleMap.OnCameraChangeListener) {
+            ((GoogleMap.OnCameraChangeListener) mAlgorithm).onCameraChange(mMap.getCameraPosition());
+        }
+
         cluster();
     }
 
+    public void setClusterOnlyVisibleArea(boolean onlyVisibleArea) {
```
markmcd: I don't think that a user should have to provide an algorithm as well as specifying this property. Can you refactor such that this property is defined on the Algorithm implementation? e.g. add a method to Algorithm like `boolean shouldReclusterOnMapMovement()`

library/src/com/google/maps/android/clustering/algo/VisibleNonHierarchicalDistanceBasedAlgorithm.java
```diff
@@ -0,0 +1,224 @@
...
+ * High level algorithm:<br>
+ * 1. Iterate over items in the order they were added (candidate clusters).<br>
+ * 2. Create a cluster with the center of the item. <br>
+ * 3. Add all items that are within a certain distance to the cluster. <br>
+ * 4. Move any items out of an existing cluster if they are closer to another cluster. <br>
+ * 5. Remove those items from the list of candidate clusters.
+ * <p/>
+ * Clusters have the center of the first element (not the centroid of the items within it).
+ */
+public class VisibleNonHierarchicalDistanceBasedAlgorithm<T extends ClusterItem>
```
markmcd: There's a lot of code duplicated here from NonHierarchicalDistanceBasedAlgorithm that you could save if you extended it. Can you do that? Feel free to loosen up the member privacy as needed.

library/src/com/google/maps/android/clustering/algo/VisibleNonHierarchicalDistanceBasedAlgorithm.java
```diff
@@ -0,0 +1,224 @@
...
+        final Map<QuadItem<T>, Double> distanceToCluster = new HashMap<QuadItem<T>, Double>();
+        final Map<QuadItem<T>, StaticCluster<T>> itemToCluster = new HashMap<QuadItem<T>, StaticCluster<T>>();
+
+        synchronized (mQuadTree) {
+
+            Bounds visibleBounds = getVisibleBounds(discreteZoom);
+
+            Collection<QuadItem<T>> items = mQuadTree.search(visibleBounds);
+
+            for (QuadItem<T> candidate : items) {
```
markmcd: As this block is straight from the NonHierarchicalDistanceBasedAlgorithm, it'd be great to pull it out into a separate method, on the parent class, when you subclass it.

barbeau:
FYI - looks like the Travis timeout issue should hopefully be fixed in `2.0.0-beta3` - see https://code.google.com/p/android/issues/detail?id=189764#c17 for details.

library/src/com/google/maps/android/clustering/algo/VisibleNonHierarchicalDistanceBasedAlgorithm.java
zamesilyasa: Sure, I didn't wanted to change initial algorithm.
Ok, I will refactor NonHierarchicalDistanceBasedAlgorithm for inheritance.

library/src/com/google/maps/android/clustering/ClusterManager.java
zamesilyasa: What about back compatibility?

markmcd: If we could use Java 8 we could add a default method to the interface, but alas that's not an option.

How about extending Algorithm (e.g. `ViewBasedAlgorithm`) and use that to determine instead. You may even be able to move the onCameraChangeListener code in there, too.

zamesilyasa: I thought about this variant too, but in this case I will have to use instanceof to check, if this algorythm is ViewBasedAlgorithm, but if this algorithm will be decorated(like you do with PreCachingAlgorithmDecorator), it will not work at all.

markmcd: Ahh, good point. My main concern is that the behaviour of using `setClusterOnlyVisibleArea` on anything other than the new Algorithm you've introduced is undefined.

The only other option available is to break the Algorithm interface by adding the method suggested above. I've thought about this a bit and I think that it's the right thing to do but I (or whoever releases the next version to Maven) will need to make sure we bump the version number appropriately so it's clear what's happening.

What do you think about that? Reasonable? Or too drastic?

larten:
I think the getVisibleBounds() method in the VisibleNonHierarchicalDistanceBasedAlgorithm class not working correctly. On bigger screens the bounds not matching to the screen and the displayed area with markers is more bigger. I've tried to fix it with getVisibleRegion() method and calculation, but if user rotate the map, this solution is not working :)
Do you have any idea?

zamesilyasa:
What screen size/dimension do you use? 
Also, if  you rotate the map, you have to recluster everything. 
There is updateViewSize() method in the latest update, call it with the new map size and recluster it.

larten:
I've checked with a 480x800 and a 1920x1080 resolution.

Now, I've solved with this code:

`private Bounds getVisibleBounds(int zoom) {
        Log.i("TAG", "mapWidth: " + zoom);
        if (mMapCenter == null) {
            return new Bounds(0, 0, 0, 0);
        }

        Point p = PROJECTION.toPoint(mMapCenter);
        Point left = PROJECTION.toPoint(this.left);
        Point right = PROJECTION.toPoint(this.right);

        return new Bounds(Math.min(left.x, right.x), Math.max(left.x, right.x),
                Math.min(left.y, right.y), Math.max(left.y, right.y));
    }

    private LatLng left;
    private LatLng right;

    @Override
    public void onCameraChange(CameraPosition cameraPosition) {
        VisibleRegion visibleRegion = map.getProjection().getVisibleRegion();
        mMapCenter = cameraPosition.target;
        left = visibleRegion.farLeft;
        right = visibleRegion.nearRight;
    }`

zamesilyasa:
I believe you use too old code. I don't use map bounds anymore. Please, pull the latest version of the patch

larten:
Okey, thank you!

meierjan:
Any idea when this is gonna be merged?

lmbd:
Is it a part of the plan to integrate / merge this functionality?

zamesilyasa:
I am just trying to keep it up to date

lfdversluis:
Is there any reason not to merge this? (besides perhaps squashing commits)

zamesilyasa:
Commits will be squashed by merge, isn't it?

lfdversluis:
Someone with merge rights could, indeed.

jeffdgr8:
@googlebot I consent to have my commit merged to this project. Looking forward to having this pull request as well as my other two that are pending merged into the base project.

b1n0m13:
@zamesilyasa I noticed that `NonHierarchicalViewBasedAlgorithm.shouldReclusterOnMapMovement()` always return `true`, causing recluster on map movement even if it is just centering at selected cluster.
So when user taps at cluster to see info window it is shown only for a second, then disappear.

zamesilyasa:
@b1n0m13 
I had this problem before, but I solved it with custom popup windows. I had a view on top of the map, which showed popup windows. It was not hard to implement, because map is centered after click. So just show a popup at the center of a layout.

> always return true
That's the core idea of the algorithm - recluster on map movement.

If you will disable this in case popup is shown and scroll the map, you will see an empty square of the map, what breaks consistency between visual representation and data set

zamesilyasa:
@markmcd 
Is there any chance to merge it into dev?

manish-bansal:
Will this be merged with main code?

stephenmcd:
This is merged now, thanks all.

manish-bansal:
Is there any plan to release this as part of next release? Managing gradle dependencies are much easier than moving the aar in multiple project.

tmiyamon:
I also want to use this feature in my project with gradle dependencies. Please release it !

lordplagus02:
+1 Would love this

Sammekl:
Could someone tell me how I can add this to my project? I am looking for an .aar or .jar file to be to use this new feature as it's still not part of a new release

PR #228, opened 2015-12-08
Allow numeric boolean parsing in kml styles

nealsanche:
Adds a KmlBoolean class that encodes the boolean parsing method so that
both 'true' and '1' will be interpreted as a boolean true value, all other values
will return false.

Modify the KmlStyleParser to use this method for parsing booleans instead of
using Boolean.parseBoolean(string).

Adds two unit tests, and sample KML data files that implement this handling.


library/src/com/google/maps/android/kml/KmlBoolean.java
```diff
@@ -0,0 +1,13 @@
+package com.google.maps.android.kml;
```
markmcd: Can you please add a copyright header to this file? Copy it from an existing file and bump the year to 2016.

nealsanche: Will do.
On Feb 18, 2016 9:03 PM, "Mark McDonald" <notifications@github.com> wrote:

> In library/src/com/google/maps/android/kml/KmlBoolean.java
> <https://github.com/googlemaps/android-maps-utils/pull/228#discussion_r53420634>
> :
>
> > @@ -0,0 +1,13 @@
> > +package com.google.maps.android.kml;
>
> Can you please add a copyright header to this file? Copy it from an
> existing file and bump the year to 2016.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/googlemaps/android-maps-utils/pull/228/files#r53420634>
> .
>

-- 
Robots and Pencils Inc | 1507 14th Ave SW Calgary AB T3C 0W4
You can unsubscribe from the Robots and Pencils mailing lists by clicking 
here. <https://robotsandpencils.com/contact/unsubscribe.html>

nealsanche: Have done. See commit below.

markmcd:
LGTM - thanks!

PR #229, opened 2015-12-09
Prevent crash on null Map Projection

apsherid:
There are race conditions that seem to be able to result in multiple map initializations causing a null projection for a tiny amount of time, this will cause several crashes if we don't properly protect ourselves from them.


apsherid:
I work for Google (apsherid@google.com), so am part of that organizations corporation.

markmcd:
LGTM. Thanks, Alan.

PR #244, opened 2016-02-17
makeIcon takes CharSequence not String

jonathan-caryl:
The underlying TextView API takes a CharSequence, so this is a simple change that lets us use different fonts etc when creating our icons. I've updated the sample app to show it in use.

sample_screenshot

### DIFF
--- a/demo/src/com/google/maps/android/utils/demo/IconGeneratorDemoActivity.java
+++ b/demo/src/com/google/maps/android/utils/demo/IconGeneratorDemoActivity.java
@@ -17,11 +17,18 @@
 package com.google.maps.android.utils.demo;
 
 import android.graphics.Color;
+import android.text.SpannableStringBuilder;
+import android.text.style.StyleSpan;
+
 import com.google.android.gms.maps.CameraUpdateFactory;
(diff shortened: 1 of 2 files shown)


jonathan-caryl:
I signed it!

jonathan-caryl:
I'm not sure from the googlebot emails whether it's happy that I've signed the CLA or not: my regular google account I use isn't the email I regularly use for github. I've now added that googlemail account, and after some faffing I've added my github email to that google account. Hopefully that's everything sorted.

markmcd:
It looks like the problem is that your Git commit uses a different email than the one that signed the CLA. At this point in time I can see that they both point to the same GitHub account, but to make future contributions smoother you can either use a different email on the commit, or add that email to your Google account so googlebot knows who you are.

That aside, contribution looks good, thanks!

jonathan-caryl:
Sweet. Any timescale for a `0.4.2` release to pick up this change?

markmcd:
There are a couple of other PRs that only need minor changes before they can be merged - I'll give them a chance to get their changes in and then release 0.4.2.

PR #245, opened 2016-02-19
Equals for cluster subclasses

hypest:
This is a re-submission of https://github.com/googlemaps/android-maps-utils/pull/169 since I no longer have commit access to that one.

For reference, here's the original PR introduction:

> Adds implementations of "equals()" and "hashCode()" in StaticCluster and QuadItem classes. Benefit is the ability to reliably compare (and identify) instances of those classes.
> 
> As a bonus (and an implicit way to test QuadItem::hashCode() without exposing the inner class) this PR also includes an implementation of NonHierarchicalDistanceBasedAlgorithm::removeItem(). By the way, this would effectively fix issue #133.
> 
> PR commits include relevant tests. Let me know if more testing should be added.
> 
> Also, the added test files are new and thus copyrighted after myself, but still under the Apache v2 license. Let me know if this needs to be revised somehow.
> 
> Cheers!

Let me know if this causes any problems. Thanks!

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -9,3 +9,4 @@
 * Mihai Preda, https://github.com/preda
 * Iris Uy, https://github.com/irisu
 * Emma Yeap, https://github.com/microcat
+* Stefanos Togoulidis, https://github.com/hypest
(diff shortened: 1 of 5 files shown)


hypest:
It seems that the build is failing in Travis but not for reasons related to the changes themselves. I think re-triggering Travis might give this a chance to succeed. @markmcd would you mind restarting the Travis build on this one? I can try to do it from my side by closing and re-opening this PR but I'll wait for your input first. Thanks!

markmcd:
Yes indeed, apologies. Builds are flaky at the moment.

Looks good - thanks again.

PR #247, opened 2016-02-20
Added the setClickable method

Mkohm:
Added the setClickable method to make the LineString clickable.

### DIFF
--- a/library/src/com/google/maps/android/geojson/GeoJsonLineStringStyle.java
+++ b/library/src/com/google/maps/android/geojson/GeoJsonLineStringStyle.java
@@ -51,10 +51,20 @@ public class GeoJsonLineStringStyle extends Observable implements GeoJsonStyle {
     }
     
     /**
-     * Sets the GeoJsonLineString clickable
-     *
+     * Gets the clickability setting for this Options object
+     * 
+     * @return true if the GeoJsonLineString is clickable; false if it is not
      */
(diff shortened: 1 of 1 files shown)


Mkohm:
I signed it!

library/src/com/google/maps/android/geojson/GeoJsonLineStringStyle.java
```diff
@@ -49,6 +49,15 @@ public void setColor(int color) {
         mPolylineOptions.color(color);
         styleChanged();
     }
+    
+    /**
+     * Sets the GeoJsonLineString clickable
```
markmcd: The JavaDoc on the others has been copied from the Android Maps API. Please do the same here - that is:
> Specifies whether this polyline is clickable.

Also remove the empty line below.

library/src/com/google/maps/android/geojson/GeoJsonLineStringStyle.java
```diff
@@ -49,6 +49,15 @@ public void setColor(int color) {
...
+     *
+     */
+    public void setClickable(boolean clickable){
```
markmcd: Can you also implement [`isClickable`](https://developers.google.com/android/reference/com/google/android/gms/maps/model/PolylineOptions.html#isClickable()) here?

PR #251, opened 2016-02-26
Increasing ADB timeout to stabalise builds

markmcd:
PTAL @jfschmakeit @domesticmouse @barbeau 

Tests look fine but I'm very slightly concerned with using an unstable version of the gradle plugin. We can always revert this if we have to but I am thoroughly sick of restarting travis builds every goddamn day, so think it's worth the risk.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -1,10 +1,11 @@
 buildscript {
     repositories {
-        mavenCentral()
+        jcenter()
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.3.0'
+        // TODO: move this to a stable release as soon as one becomes available
+        classpath 'com.android.tools.build:gradle:2.0.0-beta5'
     }
 }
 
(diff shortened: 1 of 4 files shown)


barbeau:
For reference, here's the AOSP issue that's supposedly now fixed that allows us to increase the timeout:
https://code.google.com/p/android/issues/detail?id=189764

domesticmouse:
LGTM

jfschmakeit:
LGTM

library/build.gradle
```diff
@@ -35,6 +36,13 @@ android {
             java.srcDirs = ['tests/src']
         }
     }
+
+    // This enables long timeouts required on slow environments, e.g. Travis
+    adbOptions {
```
barbeau: FWIW, I don't think this is actually needed, as I believe Travis & Gradle pick up the ABD timeout from the `ADB_INSTALL_TIMEOUT` environmental variable defined in `.travis.yml`.  So bumping Gradle plugin alone may be enough to prevent timeouts.  I don't think it hurts to have both, although I'm not sure which value wins, this timeout value or the `ADB_INSTALL_TIMEOUT`.

kevalpatel2106: I checked with sample project. In the above case the timeout value mentioned in gradle wins over ADB_INSTALL_TIMEOUT.

PR #255, opened 2016-03-07
Replace deprecated getMap with getMapAsync in demo.

stephenmcd:
Contrary to our discussion, I opted to keep the base demo class - about half of the demos are structured in a way that depends on having a map member available, so it didn't make sense to restructure everything in an async way. PTAL @markmcd @jfschmakeit

### DIFF
--- a/demo/src/com/google/maps/android/utils/demo/BaseDemoActivity.java
+++ b/demo/src/com/google/maps/android/utils/demo/BaseDemoActivity.java
@@ -19,9 +19,10 @@ package com.google.maps.android.utils.demo;
 import android.os.Bundle;
 import android.support.v4.app.FragmentActivity;
 import com.google.android.gms.maps.GoogleMap;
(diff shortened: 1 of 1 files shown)


demo/src/com/google/maps/android/utils/demo/BaseDemoActivity.java
```diff
@@ -32,23 +33,23 @@ protected int getLayoutId() {
...
-            startDemo();
-        }
+    @Override
+    public void onMapReady(GoogleMap map) {
+        mMap = map;
+        startDemo();
+    }
+
+    private void setUpMap() {
+        ((SupportMapFragment) getSupportFragmentManager().findFragmentById(R.id.map)).getMapAsync(this);
```
markmcd: You should guard this line with an if (mMap != null) check.

stephenmcd: That's done, thanks.

Issue #257, opened 2016-03-10
Polygon does not fill correctly.


SeijiIto:
I attempting to draw a filled Polygon as the same image rendered by GeoJSONLint.com.

![geojsonlint](https://cloud.githubusercontent.com/assets/9032143/13663160/067606c6-e6e3-11e5-9743-e55e909477c7.png)

I tried to do it customizing demo app.
The polygon is stroked but not filled.

![demo](https://cloud.githubusercontent.com/assets/9032143/13663298/e4894d6a-e6e3-11e5-8b63-e7e9a59ac31b.png)

My code

**GeoJsonDemoActivity.java**

```
@Override
protected void startDemo() {
    try {
        GeoJsonLayer layer = new GeoJsonLayer(getMap(), R.raw.test, getApplicationContext());
        for (GeoJsonFeature feature : layer.getFeatures()) {
            GeoJsonPolygonStyle style = new GeoJsonPolygonStyle();
            style.setFillColor(0x80ff7fff);
            style.setStrokeColor(0x80ff7fff);
            feature.setPolygonStyle(style);
        }
        layer.addLayerToMap();
    } catch (Exception e) {
        Log.e("TEST", "Exception occured.", e);
    }
}
```

**test.json**

```
{
    "type": "Feature",
    "geometry": {
        "type": "Polygon",
        "coordinates": [
            [
                [
                    139.793638,
                    35.7324
                ],
                [
                    139.793638,
                    35.733441
                ],
                [
                    139.79520000000002,
                    35.733441
                ],
                [
                    139.79520000000002,
                    35.734483
                ],
                [
                    139.79520000000002,
                    35.735525
                ],
                [
                    139.79676199999994,
                    35.735525
                ],
                [
                    139.79676199999994,
                    35.736566
                ],
                [
                    139.79832499999998,
                    35.736566
                ],
                [
                    139.79832499999998,
                    35.737608
                ],
                [
                    139.79832499999998,
                    35.738649
                ],
                [
                    139.799887,
                    35.73865
                ],
                [
                    139.799887,
                    35.737608
                ],
                [
                    139.799887,
                    35.736566
                ],
                [
                    139.799887,
                    35.735525
                ],
                [
                    139.799887,
                    35.734483
                ],
                [
                    139.799887,
                    35.733441
                ],
                [
                    139.80144999999993,
                    35.733441
                ],
                [
                    139.80144999999993,
                    35.734483
                ],
                [
                    139.80301199999997,
[...]

PR #263, opened 2016-03-18
Workaround for the issue #257

SeijiIto:
This is workaround for the issue [#257](https://github.com/googlemaps/android-maps-utils/issues/257) using [clipper-java](https://github.com/lightbringer/clipper-java) removed lambda.

Recalculate interior polygons smaller than the original holes.


stephenmcd:
https://github.com/googlemaps/android-maps-utils/issues/257#issuecomment-230195788

Issue #259, opened 2016-03-11
Demo app won't build


dant3:
`./gradlew :demo:aDebug` fails with:
```
A problem occurred configuring project ':demo'.
> failed to find Build Tools revision 23.0.0
```

but there is no such Build Tools version in Android SDK. You can only install 23.0.1/23.0.2 as closest ones

![image](https://cloud.githubusercontent.com/assets/820057/13707719/f7509990-e778-11e5-93e8-dfacf1972448.png)

ssj36:
**Error:(25, 13)** Failed to resolve: com.google.maps.android:android-maps-utils:x.y.z
Getting above error after trying rebuild project

dant3:
This is very good that this seems to be fixed since PR wich appropriate changes was merged, however, it would be very nice to actually close the issue afterwards, marking it as fixed.

Issue #252, opened 2016-03-02
Layout Names Clashes


brittBarak:
I had a layout in my project named text_bubble.xml

Trying to instantiate ClusterManager causes an exception, 
since you use a layout by the same name,

markmcd:
I agree that it would be a good idea to use a namespace prefix (e.g. amu_) for resources.

stephenmcd:
Fixed in #264

PR #264, opened 2016-03-24
Fix build issues and resource names clash

Vandalko:
Fixes:
 - https://github.com/googlemaps/android-maps-utils/issues/259
 - https://github.com/googlemaps/android-maps-utils/issues/252


Vandalko:
I signed it!

library/build.gradle
```diff
@@ -48,7 +50,7 @@ android {
 task instrumentTest(dependsOn: connectedCheck)
 
 task apklib(type: Zip) {
```
Vandalko: does anyone use apklib nowadays ? Eclipse users ?

Vandalko:
build failed because of mis-configured Travis CI
`failed to find Build Tools revision 23.0.2`

library/build.gradle
```diff
@@ -23,7 +23,9 @@ dependencies {
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "23.0.0"
+    buildToolsVersion "23.0.2"
```
barbeau: You also need to bump `ANDROID_BUILD_TOOLS_VERSION` to match in `.travis.yml` - https://github.com/googlemaps/android-maps-utils/blob/master/.travis.yml#L15.

Vandalko: @barbeau thx for the hint - all checks have passed now

build.gradle
```diff
@@ -4,9 +4,10 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.0.0-beta5'
+        classpath 'com.android.tools.build:gradle:2.1.0-alpha4'
```
markmcd: Are you using something from the latest alpha release? If not, I'd prefer to leave this dependent on the theoretically more stable beta* releases if we can't use the full releases.

library/src/com/google/maps/android/ui/RotationLayout.java
```diff
@@ -26,7 +26,7 @@
  * <p/>
  * May not work with padding.
  */
-class RotationLayout extends FrameLayout {
+public class RotationLayout extends FrameLayout {
```
markmcd: This looks unrelated to the resource prefix changes.

markmcd:
Thanks for taking a look at this. In future, feel free to send through multiple small pull requests, even if one is dependent on another. It can make reviews much easier for us.

library/src/com/google/maps/android/ui/RotationLayout.java
Vandalko: it was a lint error - since this class is used in `layout.xml` file, lint requires it to be public

build.gradle
Vandalko: @markmcd I'm not useing 2.1.0 features directly, but latest Android Studio 2.1 preview requires `2.1.0-alpha4` gradle plugin.

I can revert to `2.0.0-beta7` because it's not critical. Pls let me know if I should.

library/src/com/google/maps/android/ui/RotationLayout.java
Vandalko: check `text_bubble.xml`

markmcd: OK, SGTM

build.gradle
markmcd: Yes, please do. If you're using an unstable version of Android Studio, it will suggest an unstable version of gradle. I'd prefer to keep this as stable as possible.

Vandalko:
Downgraded Android Gradle plugin to latest beta

markmcd:
Merged in 29490ad. Thanks again!

Issue #220, opened 2015-10-27
Looks like local url in kml files not working


Matroskinas:
When I try to add groundOverlay using KML Layer I get error: " java.net.MalformedURLException: Protocol not found: "file/image.png". If I change location in kml file like this: <href>/storage/sdcard0/Download/image.png</href>, I get the same error...

gdoubleu:
Same here.

Issue #239, opened 2016-02-03
Support for local file with KML's <href>


gdoubleu:
Hi, 

I was wondering about if href's to local file (belonging to the Android app bundle) are supported. I have tried several things but always end up with a File Not Found error. In its simplest form, with the href containing <file name>.<extension>, I could not get the library to find the file when placed in res.

PR #267, opened 2016-04-01
Decode marker icon from local FS

Vandalko:
Major problem with [BitmapDescriptorFactory.fromPath(iconUrl)](https://developers.google.com/android/reference/com/google/android/gms/maps/model/BitmapDescriptorFactory.html) is that it requires **absolute** path and throws RuntimeException if file not found.

But KML files are often created with tools and by people who are not really aware of Android filesystem which gives:
```xml
		<Style id='icon-ci-1-normal'>
			<IconStyle>
				<scale>1.1</scale>
				<Icon>
					<href>images/icon-1.png</href>
				</Icon>
			</IconStyle>
			<LabelStyle>
				<scale>0.0</scale>
			</LabelStyle>
		</Style>
```

And it would simply crash at `KmlStyle.setIconUrl`.

There are two approaches to solve this:
 - Use `BitmapFactory.decodeFile(mIconUrl)` if `mIconUrl` is not really an URL, but a path on local FS
 - Try to guess where the file is placed:
```java
        private Bitmap tryDecodeLocalBitmap() {
            InputStream stream = null;
            try {
                stream = mContext.getAssets().open(mImageUrl);
                return BitmapFactory.decodeStream(stream);
            } catch (IOException assetsIOE) {
                try {
                    stream = mContext.getContentResolver().openInputStream(Uri.parse(mImageUrl));
                    return BitmapFactory.decodeStream(stream);
                } catch (IOException contentIOE) {
                    return BitmapFactory.decodeFile(mImageUrl);
                }
            } finally {
                closeSilently(stream);
            }
        }
```

The issue with second approach is that it's not really deterministic - it can pick unwanted files. That's why we stick with simple strategy - just decode local file.

P.S. fix of https://github.com/googlemaps/android-maps-utils/issues/220

### DIFF
--- a/library/src/com/google/maps/android/kml/KmlRenderer.java
+++ b/library/src/com/google/maps/android/kml/KmlRenderer.java
@@ -823,9 +823,9 @@ import java.util.Iterator;
             try {
                 return BitmapFactory.decodeStream((InputStream) new URL(mIconUrl).getContent());
             } catch (MalformedURLException e) {
-                e.printStackTrace();
+                return BitmapFactory.decodeFile(mIconUrl);
             } catch (IOException e) {
-                e.printStackTrace();
+                Log.e(LOG_TAG, "Image [" + mIconUrl + "] download issue", e);
             }
             return null;
         }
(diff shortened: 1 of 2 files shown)


Vandalko:
oh, btw, it fixes https://github.com/googlemaps/android-maps-utils/issues/220

stephenmcd:
Thanks @Vandalko 

Would it be better to check the URL scheme (http) rather than relying on exception handling for local files? Also could you add a test case?

stephenmcd:
Also closes #239

PR #279, opened 2016-05-17
Add marker-item mapping

panzerfahrer:
I always struggled with mapping my entities to the corresponding map markers. This PR adds to the existing `MarkerManager.MarkerCollection` an `MarkerManager.MarkerItemCollection`.

The `MarkerItemCollection` encapsulates the mapping of entities to map markers. This collection will be used to add and remove entities. Listeners can be registered to the collection (both `MarkerCollection` and the new `MarkerItemCollection`) to get notified of changes. Changes to the collection then reflect onto the map by re-clustering items.

Also included is partly rewrite of the `DefaultClusterRenderer`. Now all rendering of markers will happen on a background worker thread. This benefits with custom marker rendering that takes longer than the usual drawable loading.

Furthermore, only markers within the current visible bounds will be rendered and added to the map.


panzerfahrer:
I signed it!

broady:
I think this is built into the base Android Maps API now (tags), so we can close this?

barbeau:
https://developers.google.com/android/reference/com/google/android/gms/maps/model/Marker.html#setTag(java.lang.Object)

PR #286, opened 2016-06-08
Add support for click listeners on GeoJSON layers.

stephenmcd:
@markmcd PTAL, as discussed:

- Features mapped to map objects are now stored in two-way map (BiMultiMap) so we can lookup feature by map object, when it's clicked on 
- BiMultiMap also handles collection values, so each collection item is stored as a key in the reverse map
- GeoJSONLayer allows a single click listener to be created, for the common case of a single layer on a map
- Multiple layers on a map are supported via GeoJSONLater. getFeature() which can be used to determine the layer a map object belongs to, inside the developer's own map click handlers.


stephenmcd:
Build failure is related to Travis issue, all test passing locally.

demo/src/com/google/maps/android/utils/demo/GeoJsonDemoActivity.java
```diff
@@ -122,6 +122,13 @@ protected void onPostExecute(JSONObject jsonObject) {
                 // Add the layer onto the map
                 addColorsToMarkers();
                 mLayer.addLayerToMap();
+
+                mLayer.setOnFeatureClickListener(new GeoJsonLayer.GeoJsonOnFeatureClickListener() {
+                    @Override
+                    public void onFeatureClick(GeoJsonFeature feature) {
+                        Log.i(mLogTag, "Feature clicked: " + feature.getProperty("title"));
```
markmcd: Can you use a toast here instead? It'd be great to allow the user to see what's happening without having to fire up a log catcher.

library/src/com/google/maps/android/BiMultiMap.java
```diff
@@ -0,0 +1,58 @@
...
+ * for a key given a value, by storing a second HashMap internally which maps values to keys.
+ * Secondly, it supports Collection values, in which case, each item in the collection is
+ * used as a key in the internal reverse HashMap. It's therefore up to the caller to ensure
+ * the overall set of values, and collection values, are unique.
+ *
+ * Used by GeoJsonRenderer to store GeoJsonFeature instances mapped to corresponding Marker,
+ * Polyline, and Polygon map objects. We want to look these up in reverse to provide access
+ * to GeoJsonFeature instances when map objects are clicked.
+ */
+public class BiMultiMap<K> extends HashMap<K, Object> {
```
markmcd: I think this should be package private.

Also, as this extends HashMap, it should override everything relevant, e.g. clone, clear. Even if just to throw an UnsupportedOperationException. 

It also kinda bugs me that this is using Object everywhere, particularly on a public class (and later in the public methods). Have you considered using 3 separate maps and taking advantage of polymorphism to keep the interface clean?

library/src/com/google/maps/android/BiMultiMap.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private final HashMap<Object, K> mValuesToKeys = new HashMap<>();
```
markmcd: This can be declared as type Map

library/src/com/google/maps/android/BiMultiMap.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public void putAll(HashMap<K, Object> map) {
```
markmcd: Missing @override (will need some type param changes to make it work)

library/src/com/google/maps/android/BiMultiMap.java
```diff
@@ -0,0 +1,58 @@
...
+        for (Entry<K, Object> entry : map.entrySet()) {
+            put(entry.getKey(), entry.getValue());
+        }
+    }
+
+    @Override
+    public Object put(K key, Object value) {
+        // Store value/key in the reverse map, and store each item in the value
+        // if the value is a collection.
+        if (value instanceof Collection) {
```
markmcd: You could probably do this a little cleaner with a separate method. e.g.

    public Object put(K key, Collection value) ...

library/src/com/google/maps/android/BiMultiMap.java
```diff
@@ -0,0 +1,58 @@
...
+            for (Object valueItem: (Collection) value) {
```
markmcd: style nit: put a space either side of the colon.

library/src/com/google/maps/android/BiMultiMap.java
```diff
@@ -0,0 +1,58 @@
...
+                mValuesToKeys.put(valueItem, key);
+            }
+        } else {
+            mValuesToKeys.put(value, key);
+        }
+        return super.put(key, value);
+    }
+
+    public Object remove(Object key) {
```
markmcd: Missing @override

library/src/com/google/maps/android/BiMultiMap.java
```diff
@@ -0,0 +1,58 @@
...
+        // Also remove the value/key from the reverse map.
+        mValuesToKeys.remove(get(key));
```
markmcd: This won't work for collections. Perhaps add another method to handle that case?

library/src/com/google/maps/android/geojson/GeoJsonLayer.java
```diff
@@ -71,6 +74,69 @@ public GeoJsonLayer(GoogleMap map, int resourceId, Context context)
...
+     * Marker, Polyline) is clicked.
+     *
+     * Note that if multiple GeoJsonLayer objects are bound to a GoogleMap object, calling
+     * setOnFeatureClickListener on one will override the listener defined on the other. In
+     * that case, you must define each of the GoogleMap click listeners manually
+     * (OnPolygonClickListener, OnMarkerClickListener, OnPolylineClickListener), and then
+     * use the GeoJsonLayer.getFeature(mapObject) method on each GeoJsonLayer instance to
+     * determine if the given mapObject belongs to the layer.
+     *
+     * @param listener GeoJsonOnFeatureClickListener object providing the onFeatureClick
```
markmcd: You don't need to specify the type for the param in javadoc - it's apparent from the name. Please remove all through this change.

library/src/com/google/maps/android/geojson/GeoJsonLayer.java
```diff
@@ -71,6 +74,69 @@ public GeoJsonLayer(GoogleMap map, int resourceId, Context context)
...
+     * Callback interface for when a GeoJsonLayer's map object is clicked.
+     */
+    public interface GeoJsonOnFeatureClickListener {
+        void onFeatureClick(GeoJsonFeature feature);
+    }
+
+    /**
+     * Retrieves a corresponding GeoJsonFeature instance for the given map object
+     * (Polygon, Marker, Polyline). Made available so that maps with multiple layers
+     * can implement their own click listeners and determine which layer the
```
markmcd: I think this is useful beyond just click handling, so you should drop or reword to something like "Useful for maps with multiple..."

library/src/com/google/maps/android/geojson/GeoJsonLayer.java
```diff
@@ -71,6 +74,69 @@ public GeoJsonLayer(GoogleMap map, int resourceId, Context context)
...
+     * corresponding feature belongs to.
+     *
+     * @param mapObject Object map object, presumably clicked on
```
markmcd: I think the 'presumably click on' is conjecture, drop it?

library/tests/src/com/google/maps/android/BiMultiMapTest.java
```diff
@@ -0,0 +1,44 @@
...
+public class BiMultiMapTest extends TestCase {
+
+    private BiMultiMap<String> mMap = new BiMultiMap<>();
+
+    public void testSingle() {
+        String key = "foo";
+        String value = "bar";
+        mMap.clear();
+        mMap.put(key, value);
+        assertEquals(mMap.size(), 1);
```
markmcd: assert params should be expected first, then actual. This matters because a failing test shows a log statement like "Expected X, Got Y".

stephenmcd:
BTW here's the diff of what's changed since first review:

https://github.com/stephenmcd/android-maps-utils/compare/d40765ea69...251159e307

library/src/com/google/maps/android/BiMultiMap.java
```diff
@@ -16,40 +17,64 @@
  */
 public class BiMultiMap<K> extends HashMap<K, Object> {
```
markmcd: Class is still public - can you make it package private? I think you'll need to move it but that's probably fine yeah?

library/tests/src/com/google/maps/android/BiMultiMapTest.java
```diff
@@ -7,37 +7,37 @@
 
 public class BiMultiMapTest extends TestCase {
 
-    private BiMultiMap<String> mMap = new BiMultiMap<>();
-
     public void testSingle() {
+        BiMultiMap<String> mMap = new BiMultiMap<>();
```
markmcd: You should rename this, it's no longer a member so should not have the m prefix.

library/tests/src/com/google/maps/android/BiMultiMapTest.java
```diff
@@ -7,37 +7,37 @@
...
         String key = "foo";
         String value = "bar";
         mMap.clear();
```
markmcd: dd

library/tests/src/com/google/maps/android/BiMultiMapTest.java
```diff
@@ -7,37 +7,37 @@
...
+        assertEquals(0, mMap.size());
         assertNull(mMap.get(key));
         assertNull(mMap.getKey(value));
     }
 
     public void testMulti() {
+        BiMultiMap<String> mMap = new BiMultiMap<>();
         String key = "foo";
         List<String> values = Arrays.asList("bar", "baz");
         mMap.clear();
```
markmcd: dd

stephenmcd:
Last round of feedback done in 0855b264697202bf3d20134dec72f2698a9be9fc

markmcd:
LGTM

[LGTM](http://lgtm.in/i/Yegv08CXp)

PR #289, opened 2016-06-29
Null GeoJson feature properties

stephenmcd:
GeoJson can contain null values, but the parser stored them as the string "null", this fixes that.

@markmcd PTAL


markmcd:
Any tests need to be updated? Or added?

If not, LGTM

stephenmcd:
Got a test now in 1681daf566aaf6b74c2c0c3465af74ebcda944f5

markmcd:
LGTM :+1:

PR #293, opened 2016-07-11
Kml draw order

stephenmcd:
This adds support for the `drawOrder` element in KML which was originally omitted. 

@samthor PTAL


samthor:
LGTM

PR #318, opened 2016-10-03
Bump dependencies versions

barbeau:
This brings the project up-to-date with the current versions of:

* API Level
* Build Tools
* Google Maps library
* Android Gradle plugin

Travis script also modified for the above.  This allows use with Android Studio 2.2 as well.

API Level 24 emulator seems to be working, so I dropped the separate `EMULATOR_API_LEVEL` and am just using `ANDROID_API_LEVEL` for everything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: android
-jdk: oraclejdk7
+jdk: oraclejdk8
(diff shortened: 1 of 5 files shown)


stephenmcd:
Thanks Sean

PR #321, opened 2016-10-25
Reuse cluster markers in DefaultClusterRenderer

jeffdgr8:
There was a bug where if a cluster already has a marker on the map when
a cluster operation occurs, the cluster would add a new marker without
reusing the existing marker. The new marker would update the mapping to
`mClusterToMarker`, adding a new mapping in `mMarkerToCluster`, while
leaving the old marker mapped in `mMarkerToCluster`. When the old marker
is then removed, it also removes the new marker's mapping in
`mClusterToMarker`.

This results in there no longer existing a mapping for the cluster in
`mClusterToMarker` and `getMarker()` returns null. If
`OnClusterClickListener.onClusterClick()` wants to do something like this:

```
public boolean onClusterClick(Cluster<ClusterItem> cluster) {
    Marker marker = mClusterRenderer.getMarker(cluster);
    marker.showInfoWindow(); <-- NullPointerException
}
```

it is unable.

This is easily reproducible by zooming on the map such that the clusters
don't change at all but a new cluster operation occurs. All of the
clusters that did not change at all will have the described issue.

The change uses the same pattern as cluster item uses to reuse an
existing marker for a cluster if it already exists, rather than creating
a new one and removing the old one. This ensures `mClusterToMarker` and
`mMarkerToCluster` are always the same size.

Also removed the `OnInfoWindowClickListener`s in `onRemove()` the same as
`OnMarkerClickListener`s. Both are added in `onAdd()`.

Also, as `mMap.getProjection()` is annotated as `@NonNull`, the check for
`projection == null` was unneccesary.

### DIFF
--- a/library/src/com/google/maps/android/clustering/view/DefaultClusterRenderer.java
+++ b/library/src/com/google/maps/android/clustering/view/DefaultClusterRenderer.java
@@ -175,7 +175,9 @@ public class DefaultClusterRenderer<T extends ClusterItem> implements ClusterRen
     @Override
     public void onRemove() {
         mClusterManager.getMarkerCollection().setOnMarkerClickListener(null);
+        mClusterManager.getMarkerCollection().setOnInfoWindowClickListener(null);
         mClusterManager.getClusterMarkerCollection().setOnMarkerClickListener(null);
+        mClusterManager.getClusterMarkerCollection().setOnInfoWindowClickListener(null);
     }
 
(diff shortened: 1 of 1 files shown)


jeffdgr8:
I can't sign the CLA as it keeps giving me `A server error occurred, please try your request again.`

stephenmcd:
Awesome, thanks Jeff.

PR #331, opened 2016-12-01
Add new containsLocation() that takes latitude and longitude instead `LatLng`

thuytrinh:
In our project, we use different data structure to represent coordinates rather than `LatLng`. As a result, in order to utilize `containsLocation()`, we have to convert into `LatLng` which causes an allocation. So if we call it inside a loop through a list of coordinates, that will cause a lot more allocations.

For example:
```java
for (Location location : locations) {
  if (PolyUtil.containsLocation(new LatLng(location.lat(), location.lng()), polygon, true)) {
    return true;
  }
}
```

So it'd be better if we don't allocate `LatLng`s:
```java
for (Location location : locations) {
  if (PolyUtil.containsLocation(location.lat(), location.lng(), polygon, true)) {
    return true;
  }
}
```

### DIFF
--- a/library/src/com/google/maps/android/PolyUtil.java
+++ b/library/src/com/google/maps/android/PolyUtil.java
@@ -87,6 +87,10 @@ public class PolyUtil {
             mercator(lat3) >= mercatorLatRhumb(lat1, lat2, lng2, lng3);
     }
 
+    public static boolean containsLocation(LatLng point, List<LatLng> polygon, boolean geodesic) {
+        return containsLocation(point.latitude, point.longitude, polygon, geodesic);
+    }
+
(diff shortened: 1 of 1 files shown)


thuytrinh:
I signed it!

thuytrinh:
I signed it!

barbeau:
+1 for this change, although we'll need to run unit tests manually - there seems to be an issue with recent API Level 24 emulator so Travis builds are hanging (for other projects too - see https://github.com/travis-ci/travis-ci/issues/6122#issuecomment-263713701).

stephenmcd:
Thanks @thuytrinh!

PR #339, opened 2016-12-15
Configurable max distance between clustered items

jeffdgr8:
Added a setter/getter for the max distance between clustered items in the clustering Algorithm interface and implemented in NonHierarchicalDistanceBasedAlgorithm and GridBasedAlgorithm.


stephenmcd:
Thanks again!

PR #351, opened 2017-01-25
GeoJSON/KML Integration

Libby713:


stephenmcd:
Hey @barbeau, just wanted to give you a chance to have a look at this and give any feedback if you have time. If not then no problem, as @jfschmakeit and I have already reviewed this internally, but you might pick up something we missed! 

Here's a brief overview of the change. Previously the KML and GeoJSON implementations were quite separate even though they shared a similar design and set of features. The changes here refactors this so that the KML and GeoJSON implementations share a common base abstraction. This should make implementing features and fixes for both a lot more easier in the future. For example, last year I implemented click handling for GeoJSON, but not KML, given the previous design, but that's now been addressed as part of this new work.

We'll look at merging this early next week, and then pushing the new 0.5 release a couple of days after that. Just wanted to give you a chance to have a look before then, but again if you're pressed for time, no problem. Thanks!

barbeau:
@Libby713 and @stephenmcd Thanks for the heads up, and nice work!  Definitely sounds like a useful addition to the project.  I'll see if I can give it a quick review in the next few days, but if you don't hear from me by early next week feel free to merge - I don't want to hold things up.

demo/src/com/google/maps/android/utils/demo/GeoJsonDemoActivity.java
```diff
@@ -45,9 +46,9 @@ protected int getLayoutId() {
     @Override
     protected void startDemo() {
         // Download the GeoJSON file.
-        retrieveFileFromUrl();
+        //retrieveFileFromUrl();
```
barbeau: Is this comment out intended?  If so, I'd suggest adding a comment saying why.

library/src/com/google/maps/android/data/DataPolygon.java
```diff
@@ -0,0 +1,45 @@
...
+ * @param <T> the type of Polygon - GeoJsonPolygon or KmlPolygon
+ */
+public interface DataPolygon<T> extends Geometry {
+
+    /**
+     * Gets an array of outer boundary coordinates
+     *
+     * @return array of outer boundary coordinates
+     */
+    public ArrayList<LatLng> getOuterBoundaryCoordinates();
```
barbeau: I believe this can be relaxed to just `List`?

library/src/com/google/maps/android/data/DataPolygon.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Gets an array of arrays of inner boundary coordinates
+     *
+     * @return array of arrays of inner boundary coordinates
+     */
+    public ArrayList<ArrayList<LatLng>> getInnerBoundaryCoordinates();
```
barbeau: I believe this can be relaxed to just `List`?

library/src/com/google/maps/android/data/Feature.java
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * An abstraction that shares the common properties of
+ * {@link com.google.maps.android.data.kml.KmlPlacemark KmlPlacemark} and
+ * {@link com.google.maps.android.data.geojson.GeoJsonFeature GeoJsonFeature}
+ */
+public class Feature extends Observable {
+
+    private final String mId;
+
+    private final HashMap<String, String> mProperties;
```
barbeau: I believe this can be relaxed to just `Map`?

library/src/com/google/maps/android/data/Feature.java
```diff
@@ -0,0 +1,153 @@
...
+
+    private Geometry mGeometry;
+
+    /**
+     * Creates a new Feature object
+     * @param featureGeometry type of geometry to assign to the feature
+     * @param id common identifier of the feature
+     * @param properties hashmap containing properties related to the feature
+     */
+    public Feature(Geometry featureGeometry, String id,
```
barbeau: I believe this can be relaxed to just `Map`?

library/src/com/google/maps/android/data/MultiGeometry.java
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Creates a new MultiGeometry object
+     * @param geometries contains list of Polygons, Linestrings or Points
+     */
+    public MultiGeometry(List<?> geometries) {
+        if (geometries == null) {
+            throw new IllegalArgumentException("Geometries cannot be null");
+        }
+
+        //convert unknown geometry type (due to GeoJSON types) to Geometry type
```
barbeau: Nit - should be a space following `//` - `// Convert...`

library/src/com/google/maps/android/data/geojson/GeoJsonLineString.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Gets the coordinates of the GeoJsonLineString
+     *
+     * @return list of coordinates of the GeoJsonLineString
+     */
+    public List<LatLng> getCoordinates() {
+        return getGeometryObject();
+    }
+
```
barbeau: Nit - extra new line.

library/src/com/google/maps/android/data/geojson/GeoJsonMultiLineString.java
```diff
@@ -0,0 +1,49 @@
...
+        return getGeometryType();
+    }
+
+    /**
+     * Gets a list of GeoJsonLineStrings
+     *
+     * @return list of GeoJsonLineStrings
+     */
+    public List<GeoJsonLineString> getLineStrings() {
+        //convert list of Geometry types to list of GeoJsonLineString types
```
barbeau: Nit - should be a space after `//` - `// Convert..`.

demo/src/com/google/maps/android/utils/demo/GeoJsonDemoActivity.java
```diff
@@ -45,9 +46,9 @@ protected int getLayoutId() {
     @Override
     protected void startDemo() {
```
barbeau: The demo currently loads with the default `(0,0)` as the map center (left screenshot below).  This gives the appearance of an empty map.  I'd suggest recentering the map so you can see some of the markers on startup (for example, right screenshot):

image

library/src/com/google/maps/android/data/geojson/GeoJsonRenderer.java
```diff
@@ -0,0 +1,136 @@
...
+ */
+public class GeoJsonRenderer extends Renderer implements Observer {
+
+    private final static Object FEATURE_NOT_ON_MAP = null;
+
+    /**
+     * Creates a new GeoJsonRender object
+     *
+     * @param map map to place GeoJsonFeature objects on
+     * @param features
```
barbeau: Missing Javadoc

library/src/com/google/maps/android/data/geojson/GeoJsonLayer.java
```diff
@@ -0,0 +1,164 @@
...
+ * To remove the rendered data from the layer
+ * {@code layer.removeLayerFromMap();}
+ */
+public class GeoJsonLayer extends Layer {
+
+    private final GeoJsonRenderer mRenderer;
+
+    private LatLngBounds mBoundingBox;
+
+    private GoogleMap mMap;
```
barbeau: Looks like `mMap` isn't used and can be removed.  Or, it should be assigned in the contructors.

library/src/com/google/maps/android/data/geojson/GeoJsonLayer.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+        mBoundingBox = null;
+        GeoJsonParser parser = new GeoJsonParser(geoJsonFile);
+        // Assign GeoJSON bounding box for FeatureCollection
+        mBoundingBox = parser.getBoundingBox();
+        HashMap<GeoJsonFeature, Object> geoJsonFeatures = new HashMap<>();
+        for (GeoJsonFeature feature : parser.getFeatures()) {
+            geoJsonFeatures.put(feature, null);
+        }
+        mRenderer = new GeoJsonRenderer(map, geoJsonFeatures);
```
barbeau: I believe `mRenderer` should be converted to a local variable (instead of member variable) here?  `GeoJsonLayer` extends `Layer`, and there is a reference to the `GeoJsonLayer` superclass `Renderer` within `Layers.storeRenderer()`.

library/src/com/google/maps/android/data/kml/KmlLayer.java
```diff
@@ -50,7 +50,9 @@ public KmlLayer(GoogleMap map, InputStream stream, Context context)
         stream.close();
         mRenderer.storeKmlData(parser.getStyles(), parser.getStyleMaps(), parser.getPlacemarks(),
                 parser.getContainers(), parser.getGroundOverlays());
+        storeRenderer(mRenderer);
```
barbeau: Same comment as `geoJsonRenderer` - I believe `mRenderer` should be converted to a local variable (instead of member variable) here?  `KmlLayer` extends `Layer`, and there is a reference to the `KmlRenderer` superclass `Renderer` within `Layers.storeRenderer()`.

library/src/com/google/maps/android/data/kml/KmlLineString.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    /**
+     * Gets the coordinates
+     *
+     * @return ArrayList of LatLng
+     */
+    public ArrayList<LatLng> getGeometryObject() {
+        List<LatLng> coordinatesList = super.getGeometryObject();
+        ArrayList<LatLng> coordinatesAList = new ArrayList<>(coordinatesList);
```
barbeau: Can be simplified to `return new ArrayList<>(coordinatesList);`

library/src/com/google/maps/android/data/kml/KmlMultiGeometry.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /**
+     * Gets an ArrayList of Geometry objects
+     *
+     * @return ArrayList of Geometry objects
+     */
+    public ArrayList<Geometry> getGeometryObject() {
+        List<Geometry> geometriesList = super.getGeometryObject();
+        ArrayList<Geometry> geometriesAList = new ArrayList<>(geometriesList);
```
barbeau: Can be simplified to `return new ArrayList<>(geometriesList);`

library/src/com/google/maps/android/data/kml/KmlRenderer.java
```diff
@@ -0,0 +1,524 @@
...
+    }
+
+    /**
+     * Removes all given KML placemarks from the map and clears all stored placemarks.
+     *
+     * @param placemarks placemarks to remove
+     */
+    private void removePlacemarks(HashMap<? extends Feature, Object> placemarks) {
+        // Remove map object from the map
+        super.removeFeatures((HashMap<Feature, Object>) placemarks);
```
barbeau: `super` reference can be removed here, as `removeFeatures()` is static.

library/src/com/google/maps/android/data/geojson/GeoJsonLayer.java
```diff
@@ -0,0 +1,164 @@
...
+        StringBuilder result = new StringBuilder();
+        // Reads from stream
+        BufferedReader reader = new BufferedReader(new InputStreamReader(stream));
+        try {
+            // Read each line of the GeoJSON file into a string
+            while ((line = reader.readLine()) != null) {
+                result.append(line);
+            }
+        } finally {
+            reader.close();
```
barbeau: This can potentially throw an `IOException`.

library/src/com/google/maps/android/data/MultiGeometry.java
```diff
@@ -0,0 +1,101 @@
...
+
+    private String geometryType = "MultiGeometry";
+
+    private List<Geometry> mGeometries;
+
+    /**
+     * Creates a new MultiGeometry object
+     * @param geometries contains list of Polygons, Linestrings or Points
+     */
+    public MultiGeometry(List<?> geometries) {
```
barbeau: Unless I'm missing something, from a quick look through the uses of this class it seems like objects passed in must extend `Geometry`.  IMHO a better approach for the implementation here, assuming we want to copy the provided list and not just keep a reference to it, would be:

~~~
    public MultiGeometry(List<? extends Geometry> geometries) {
        if (geometries == null) {
            throw new IllegalArgumentException("Geometries cannot be null");
        }
        mGeometries = new ArrayList<>(geometries);
    }
~~~

Simpler, doesn't assume objects extend `Geometry`, and allows platform to optimize the copying of the `ArrayList`.

library/src/com/google/maps/android/data/kml/KmlRenderer.java
```diff
@@ -0,0 +1,524 @@
...
+    private void addGroundOverlayToMap(String groundOverlayUrl,
+            HashMap<KmlGroundOverlay, GroundOverlay> groundOverlays, boolean containerVisibility) {
+        BitmapDescriptor groundOverlayBitmap = BitmapDescriptorFactory
+                .fromBitmap(getImagesCache().get(groundOverlayUrl));
+        for (KmlGroundOverlay kmlGroundOverlay : groundOverlays.keySet()) {
+            if (kmlGroundOverlay.getImageUrl().equals(groundOverlayUrl)) {
+                GroundOverlayOptions groundOverlayOptions = kmlGroundOverlay.getGroundOverlayOptions()
+                        .image(groundOverlayBitmap);
+                GroundOverlay mapGroundOverlay = attachGroundOverlay(groundOverlayOptions);
+                if (containerVisibility == false) {
```
barbeau: Nit - for code style, to match other places this should be `!containerVisibility`.

barbeau:
Managed to give it a quick run-through - looks good!  Some comments in-line.  I'll try to dive in deeper if time allows, but again, if you don't hear more before next week feel free to merge.

barbeau:
Oh, and one last item.

From https://medium.com/google-developers/developing-for-android-ii-bb9a51f8c8b9 (written by Android team):

> Android collections classes: Unless you require a map for a large set of items, consider using [ArrayMap](https://developer.android.com/reference/android/util/ArrayMap.html) or [SimpleArrayMap](https://developer.android.com/reference/android/support/v4/util/SimpleArrayMap.html) for data structures instead of HashMap. These classes are optimized to be more memory efficient and less GC-heavy than HashMap, better matching common use cases for such structures on mobile. (As a bonus, they also support iterating over entries without an Iterator.) Also, consider setting the initial capacity of your collection to avoid automatic resizing as it grows to the appropriate size.

It'd been on my TODO list for a while to audit my own use of `HashMaps` in apps I maintain.  You might want to think about the above for android-maps-utils here, although my feeling is that use of objects in our context may get into the `large set` scale, in which case `HashMap` would still be preferable.

stephenmcd:
CLA not required, these are all Google commits.

barbeau:
:+1:

PR #360, opened 2017-02-21
Improve PolyUtil.isLocationOnPath

noberasco:
The methods provided to know whether a given location lies on the path of a polyline are good, but they could be even better if they also told *where* on the polyline the location is.

This in turn allows, for instance, to tap on any point on a polyline and get statistical information about the tapped point:
screenshot_20170222-000613-640


noberasco:
I signed it!

PR #361, opened 2017-02-21
Improve PolyUtil.isLocationOnPath to also tell where the location is on the polyline

noberasco:
The methods provided to know whether a given location lies on the path of a polyline are good, but they could be even better if they also told where on the polyline the location is.

This in turn allows, for instance, to tap on any point on a polyline and get statistical information about the tapped point:
screenshot_20170222-000613-640

### DIFF
--- a/library/src/com/google/maps/android/PolyUtil.java
+++ b/library/src/com/google/maps/android/PolyUtil.java
@@ -178,20 +178,22 @@ public class PolyUtil {
     }
 
     /**
-     * Computes whether (and where) the given point lies on or near a polyline, within a specified
-     * tolerance in meters. The polyline is composed of great circle segments if geodesic
-     * is true, and of Rhumb segments otherwise. The polyline is not closed -- the closing
-     * segment between the first point and the last point is not included.
-     * Return values:
-     * 0 if point is between poly[0] and poly[1] (inclusive),
-     * 1 if between poly[1] and poly[2],
-     * ...,
-     * poly.size()-2 if between poly[poly.size() - 2] and poly[poly.size() - 1]
-     * -1 if point does not lie on or near the polyline.
+     * Computes whether (and where) a given point lies on or near a polyline, within a specified tolerance.
+     * The polyline is not closed -- the closing segment between the first point and the last point is not included.
+     * @param point our needle
+     * @param poly our haystack
+     * @param geodesic the polyline is composed of great circle segments if geodesic
+     *                 is true, and of Rhumb segments otherwise
+     * @param tolerance tolerance (in meters)
+     * @return -1 if point does not lie on or near the polyline.
+     *          0 if point is between poly[0] and poly[1] (inclusive),
+     *          1 if between poly[1] and poly[2],
+     *          ...,
+     *          poly.size()-2 if between poly[poly.size() - 2] and poly[poly.size() - 1]
      */
-    public static int locationIndexOnPath(LatLng point, List<LatLng> polyline,
+    public static int locationIndexOnPath(LatLng point, List<LatLng> poly,
                                            boolean geodesic, double tolerance) {
-        return locationIndexOnEdgeOrPath(point, polyline, false, geodesic, tolerance);
+        return locationIndexOnEdgeOrPath(point, poly, false, geodesic, tolerance);
     }
 
     /**
@@ -205,16 +207,19 @@ public class PolyUtil {
     }
 
     /**
-     * Computes whether (and where) the given point lies on or near a polyline, within a specified
-     * tolerance in meters. The polyline is composed of great circle segments if geodesic
-     * is true, and of Rhumb segments otherwise.
+     * Computes whether (and where) a given point lies on or near a polyline, within a specified tolerance.
      * If closed, the closing segment between the last and first points of the polyline is not considered.
-     * Return values:
-     * -1 if point does not lie on or near the polyline.
-     * 0 if point is between poly[0] and poly[1] (inclusive),
-     * 1 if between poly[1] and poly[2],
-     * ...,
-     * poly.size()-2 if between poly[poly.size() - 2] and poly[poly.size() - 1]
+     * @param point our needle
+     * @param poly our haystack
+     * @param closed whether the polyline should be considered closed by a segment connecting the last point back to the first one
+     * @param geodesic the polyline is composed of great circle segments if geodesic
+     *                 is true, and of Rhumb segments otherwise
+     * @param toleranceEarth tolerance (in meters)
+     * @return -1 if point does not lie on or near the polyline.
+     *          0 if point is between poly[0] and poly[1] (inclusive),
+     *          1 if between poly[1] and poly[2],
+     *          ...,
+     *          poly.size()-2 if between poly[poly.size() - 2] and poly[poly.size() - 1]
      */
     private static int locationIndexOnEdgeOrPath(LatLng point, List<LatLng> poly, boolean closed,
                                           boolean geodesic, double toleranceEarth) {
--- a/library/src/com/google/maps/android/PolyUtil.java
+++ b/library/src/com/google/maps/android/PolyUtil.java
@@ -178,10 +178,12 @@ public class PolyUtil {
     }
 
     /**
-     * Computes whether (add where) the given point lies on or near a polyline, within a specified
+     * Computes whether (and where) the given point lies on or near a polyline, within a specified
      * tolerance in meters. The polyline is composed of great circle segments if geodesic
      * is true, and of Rhumb segments otherwise. The polyline is not closed -- the closing
      * segment between the first point and the last point is not included.
+     * Returns 0 if point is between poly[0] and poly[1], 1 if between poly[1] and poly[2], etc.
+     * Returns a negative value if point does not lie on or near the polyline.
      */
     public static int locationIndexOnPath(LatLng point, List<LatLng> polyline,
                                            boolean geodesic, double tolerance) {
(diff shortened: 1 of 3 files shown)


noberasco:
It seems travis fails because the emulator doesn't respond in time...

barbeau:
@noberasco thanks!  Don't worry about Travis, the emulator images are messed up so it currently always fails. Could you please add a unit test for the methods that take the index as a parameter?

barbeau:
Sorry, I meant methods that return index.

library/src/com/google/maps/android/PolyUtil.java
```diff
@@ -183,15 +211,17 @@ private static boolean isLocationOnEdgeOrPath(LatLng point, List<LatLng> poly, b
...
         double lat1 = toRadians(prev.latitude);
         double lng1 = toRadians(prev.longitude);
+        int idx = 0;
         if (geodesic) {
             for (LatLng point2 : poly) {
                 double lat2 = toRadians(point2.latitude);
                 double lng2 = toRadians(point2.longitude);
                 if (isOnSegmentGC(lat1, lng1, lat2, lng2, lat3, lng3, havTolerance)) {
-                    return true;
+                    return idx;
```
preda: It may be useful to clearly state the meaning of a return of 0 (idx==0), notably for closed==true. It may be more intuitive to map the "closing" segment on idx==n-1 instead of idx==0 as is now.

library/src/com/google/maps/android/PolyUtil.java
```diff
@@ -172,9 +172,37 @@ public static boolean isLocationOnPath(LatLng point, List<LatLng> polyline,
...
+                                           boolean geodesic, double tolerance) {
+        return locationIndexOnEdgeOrPath(point, polyline, false, geodesic, tolerance);
+    }
+
+    /**
+     * Same as {@link #locationIndexOnPath(LatLng, List, boolean, double)}
+     *
+     * with a default tolerance of 0.1 meters.
+     */
+    public static int locationIndexOnPath(LatLng point, List<LatLng> polyline,
```
preda: Please document the returned "int"; mention the range and in which situations negative values are returned.

library/src/com/google/maps/android/PolyUtil.java
```diff
@@ -172,9 +172,37 @@ public static boolean isLocationOnPath(LatLng point, List<LatLng> polyline,
...
+        return (idx >= 0);
+    }
+
+    /**
+     * Computes whether (add where) the given point lies on or near a polyline, within a specified
+     * tolerance in meters. The polyline is composed of great circle segments if geodesic
+     * is true, and of Rhumb segments otherwise. The polyline is not closed -- the closing
+     * segment between the first point and the last point is not included.
+     */
+    public static int locationIndexOnPath(LatLng point, List<LatLng> polyline,
```
preda: Please document the returned int.

library/src/com/google/maps/android/PolyUtil.java
```diff
@@ -172,9 +172,37 @@ public static boolean isLocationOnPath(LatLng point, List<LatLng> polyline,
...
+     * Same as {@link #locationIndexOnPath(LatLng, List, boolean, double)}
+     *
+     * with a default tolerance of 0.1 meters.
+     */
+    public static int locationIndexOnPath(LatLng point, List<LatLng> polyline,
+                                           boolean geodesic) {
+        return locationIndexOnPath(point, polyline, geodesic, DEFAULT_TOLERANCE);
+    }
+
+    private static int locationIndexOnEdgeOrPath(LatLng point, List<LatLng> poly, boolean closed,
```
preda: You may want to document here the interaction of "closed" with the returned value.

I think a useful return could be:
returns 0 if "point" is between poly[0] and poly[1], 1 if between poly[1] and poly[2], etc. If closed==true, returns N-1 if point is between poly[n-1] and poly[0]. -- but the current implementation does not achieve this.

library/src/com/google/maps/android/PolyUtil.java
noberasco: OK, will do.

library/src/com/google/maps/android/PolyUtil.java
noberasco: OK, will do.

library/src/com/google/maps/android/PolyUtil.java
noberasco: In my opinion, it would make more sense to return N if the point is between poly[n-1] and poly[0]. This would allow to distinguish between:
- point is between poly[n-2] and poly[n-1]
- point is between poly[n-1] and poly[0]

library/src/com/google/maps/android/PolyUtil.java
noberasco: See my comment above: tap on the closing segment ==> return N

library/src/com/google/maps/android/PolyUtil.java
preda: Yes; but if size(polyline)==N, and you number the segments starting at 0, then the last segment (between poly[n-2] and poly[n-1]) is N-2; thus the segment "after last" is N-1. (i.e., I was taking N to be the number of points of the polyline, not the number of segments). Otherwise I think we agree.

noberasco:
As requested I:
- added documentation on the returned int values
- added unit tests for the new methods

Additionally, I:
- fixed a compilation issue on KmlPolygonTest.java (possibly my java compiler is more up to date than yours?)

For the time being, I did not change the logic for the corner case 'closed==true and point is on the return segment'. I'm waiting for your feedback on my own counter-proposal. However, it has to be said that the new public methods locationIndexOnPath always pass 'false' to closed, so the whole point is kind of moot.

library/tests/src/com/google/maps/android/PolyUtilTest.java
```diff
@@ -125,6 +135,28 @@ public void testOnEdge() {
...
+        // Empty.
+        locationIndexCase(makeList(), new LatLng(0, 0), -1);
+
+        // One point.
+        locationIndexCase(makeList(1, 2), new LatLng(1, 2), 0);
+        locationIndexCase(makeList(1, 2), new LatLng(3, 5), -1);
+
+        // Two points.
+        locationIndexCase(makeList(1, 2, 3, 5), new LatLng(1, 2), 0);
+        locationIndexCase(makeList(1, 2, 3, 5), new LatLng(3, 5), 1);
```
preda: Probably here we should have here a return of 0, because (3,5) is considered part of the first segment [(1,2), (3,5)]. The 1 you get is a side-effect of the current implem, which considers [(1,2), (1,2)] as segment0 and [(1,2), (3,5)] as segment1.

preda:
I agree to leave or change the "closed" case as you see fit, just document
the behavior (of the private method) so that the case with closed=true is
clear (e.g. for a future developer).

Please fix the test case and implementation so that, if polyline = [A, B,
C] (3 points),
locationIndexOnPath(A, polyline) == 0,
locationIndexOnPath(in-between-A-B, polyline) == 0,
locationIndexOnPath(B, polyline) == 0,
locationIndexOnPath(in-between-B-C, polyline) == 1.

noberasco:
Done!

library/tests/src/com/google/maps/android/data/kml/KmlPolygonTest.java
```diff
@@ -5,6 +5,7 @@
 import junit.framework.TestCase;
 
 import java.util.ArrayList;
+import java.util.List;
 
 public class KmlPolygonTest extends TestCase {
```
barbeau: I believe these changes to `KmlPolygonTest` are unrelated to adding the ability to return the index for the closest point on the line.  If so, can you please remove these changes from this changeset?

library/src/com/google/maps/android/PolyUtil.java
```diff
@@ -172,9 +172,55 @@ public static boolean isLocationOnPath(LatLng point, List<LatLng> polyline,
 
     private static boolean isLocationOnEdgeOrPath(LatLng point, List<LatLng> poly, boolean closed,
                                                   boolean geodesic, double toleranceEarth) {
+        int idx = locationIndexOnEdgeOrPath(point, poly, closed, geodesic, toleranceEarth);
+
+        return (idx >= 0);
+    }
+
+    /**
```
barbeau: Could you please change the format here to include the `@param` and `@return` fields in the comments so the Javadocs generate correctly?  See `simplify()` for an example.  On a related note, the `poly` name in the comments doesn't match the name of any method parameters (I assume this should be `polyline`).

library/tests/src/com/google/maps/android/data/kml/KmlPolygonTest.java
noberasco: They fix a compilation issue I have on my side (possibly my java compiler is more up to date than yours?). Will file them as a separate pull request.

barbeau: Try rebasing this on master branch - there were some changes to the KML package name and method return types in the last release.  That should fix it.

noberasco:
OK, this should address the latest changes requested:
- @preda rationalize test cases
- @barbeau revert compilation issue patch
- @barbeau change comment format

preda:
Looks good to me, thanks!

barbeau:
@noberasco Thanks, this looks good to me!

@stephenmcd Do you want to be the one to merge?

@noberasco BTW, you're right on the `KmlPolygonTest` test compilation issue - looks like this must have snuck in at the end of https://github.com/googlemaps/android-maps-utils/pull/351 just prior to merge.  Could you please open a separate PR for that?

stephenmcd:
Thanks everyone!

PR #365, opened 2017-02-27
Fix null overlays NPE in KmlRenderer.removeGroundOverlays method

lexi-sr:


lexi-sr:
I signed it

lexi-sr:
Fixed in https://github.com/googlemaps/android-maps-utils/pull/375

PR #366, opened 2017-02-27
Relative urls directory

lexi-sr:
Allows you to configure a directory to look for relative image URLs if the KML file has relative image paths. (This helps the library be compatible with KMZs which have the images in the same folder as the KML file, and not all images are simply online).


lexi-sr:
These changes have been included in https://github.com/googlemaps/android-maps-utils/pull/374 so we only need to make one constructor for both of the changes. This can keep the code simpler from using too many telescoping constructors. I considered using the Builder pattern, but thought it would not be backwards-compatible.

PR #367, opened 2017-02-27
Fix Geometry NPE

lexi-sr:
`placemark.getGeometry().getGeometryType()` was generating a null pointer exception and crashing the app because Geometry was null, so I added a simple null check.


lexi-sr:
Fixed in https://github.com/googlemaps/android-maps-utils/pull/375/files

PR #368, opened 2017-02-27
Optimized quad tree and cluster algorithm with HashSet instead of ArrayList

jeffdgr8:
ArrayList.remove() is a much more expensive operation than HashSet.remove(). Neither of these collections need to be an array, as neither order nor duplicate entries are required. A set collection is much more optimal with very noticeable speed improvements with large sets of cluster items.

### DIFF
--- a/library/src/com/google/maps/android/clustering/algo/NonHierarchicalDistanceBasedAlgorithm.java
+++ b/library/src/com/google/maps/android/clustering/algo/NonHierarchicalDistanceBasedAlgorithm.java
@@ -52,7 +52,7 @@ public class NonHierarchicalDistanceBasedAlgorithm<T extends ClusterItem> implem
     /**
      * Any modifications should be synchronized on mQuadTree.
      */
-    private final Collection<QuadItem<T>> mItems = new ArrayList<QuadItem<T>>();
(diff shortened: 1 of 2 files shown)


barbeau:
Huge +1 for this as long as it doesn't affect algorithm correctness.  It changes time complexity from O(n) to O(1) for remove() operation (assuming `HashSet` is implemented optimally by the platform).

I've run the tests locally and they all pass, and while it's been a while since I've done a deep dive on QuadTrees, I don't believe changing from a `List` to a `Set` has any adverse side-effects.

I'd like to get another set of eyes on this that's familiar with QuadTrees before merging though.

broady:
LGTM, but might want to use TreeSet instead. I can't recall if iteration order ever matters.

barbeau:
From a quick look through the code it doesn't appear that order matters - only operations used are add and remove with simple iterations though collection.

jeffdgr8:
There's a very noticeable speed improvement in our use case where we have potentially 10,000+ markers on the map at times. When toggling them on/off the ArrayList.remove() operation was the clear bottleneck.

AAverin:
Sad, such a nice PR not merged because emulator couldn't start on travis

stephenmcd:
Thanks a lot

PR #374, opened 2017-03-13
Relative URLs & ability to optionally not set an info window adapter

lexi-sr:
`Renderer.createInfoWindow()` was calling `mMap.setInfoWindowAdapter`, overriding any custom info window adapter that the user might have set. I added the ability to construct a KmlLayer with a boolean parameter that will tell the library to not set that default info window adapter.

Also, I have the Relative URLs feature in this pull request so we do not need so many telescoping constructors. So the new KmlLayer constructor also accepts a string for a directory name. The library was modified to first look online for the image, and if it was not found, then it looks in the directory for the image.  If it is null, it will behave as before (only looking online). This feature allows users to open KMZ (zipped) folders that have offline images.

Everything was made backwards compatible by leaving the original constructors intact.


lexi-sr:
This was merging from the wrong fork. It was a mistake. The corrected pull request is here: https://github.com/googlemaps/android-maps-utils/pull/375

PR #375, opened 2017-03-13
Relative URLs & ability to optionally not set an info window adapter

lexi-sr:
`Renderer.createInfoWindow()` was calling `mMap.setInfoWindowAdapter`, overriding any custom info window adapter that the user might have set. I added the ability to construct a KmlLayer with a boolean parameter that will tell the library to not set that default info window adapter.

Also, I have the Relative URLs feature in this pull request so we do not need so many telescoping constructors. So the new KmlLayer constructor also accepts a string for a directory name. The library was modified to first look online for the image, and if it was not found, then it looks in the directory for the image.  If it is null, it will behave as before (only looking online). This feature allows users to open KMZ (zipped) folders that have offline images.

Everything was made backwards compatible by leaving the original constructors intact.

Other small fixes included in this pull request:
- `placemark.getGeometry().getGeometryType()` was generating a null pointer exception and crashing the app because Geometry was null, so I added a simple null check.
- Fix null overlays NPE in `KmlRenderer.removeGroundOverlays` method.
- adds auto-scaling of bitmaps according to screen density (see usages of `mBitmapOptions` in `KmlRenderer.java` for more info).

### DIFF
--- a/REALEASENOTES.md
+++ b/REALEASENOTES.md
@@ -1,0 +1,47 @@
+# Google Maps Android API Utility Library Release Notes
+
+### Version 0.5
+
+#### New Features:
+
+  * Click handling was previously only available for GeoJSON, but is now available for KML layers as well
+    - Also works for multipolygons, multilinestring and multipoints
+  * Info windows can be added to clustered markers
+
+
+#### Modifications from previous version:
+ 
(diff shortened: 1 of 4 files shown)


lexi-sr:
Need to make adjustments. Closing temporarily while I improve it.

chitgoks:
did you manage to finish this?

sgallego:
Any news about this PR?

PR #378, opened 2017-03-22
Handle 3xx redirects from HTTP to HTTPS

jeffdgr8:
and vice versa. HttpURLConnection does not follow these redirects natively (https://bugs.openjdk.java.net/browse/JDK-4620571). KMLs with image assets that redirect are unable to download.


stephenmcd:
Thanks again Jeff

PR #380, opened 2017-03-28
Add MapObjectManager for multi-layer support with KML/GeoJSON renderers

jeffdgr8:
MarkerManager, PolygonManager, PolylineManager, GroundOverlayManager, and CircleManager used to create collections for map objects in KML and GeoJSON renderers. This allows them to set up listeners for these map objects while also allowing other collections of any of these map objects to be used external to the KML/GeoJSON objects on the map, which can implement their own listeners through their own collections.

MapObjectManager is an abstract base class based on MarkerManger, now used for MarkerManager and the newly implemented PolygonManager, PolylineManager, GroundOverlayManager, and CircleManager.

Fix https://github.com/googlemaps/android-maps-utils/issues/459, Fix https://github.com/googlemaps/android-maps-utils/issues/420, Fix https://github.com/googlemaps/android-maps-utils/issues/430, Fix https://github.com/googlemaps/android-maps-utils/issues/594, Fix https://github.com/googlemaps/android-maps-utils/issues/538


jeffdgr8:
Merged latest changes from master and updated so it's mergeable again.

jeffdgr8:
These changes are important in order to be able to use the KML functionality alongside other usages of markers on the map, including clustering. The same applies for polygons, polylines, ground overlays, and circles. This allows each unique use case for each of these map objects to to be able to maintain separate listeners.

barbeau:
@jeffdgr8 Now that the retain bitmap cache (https://github.com/googlemaps/android-maps-utils/pull/381) PR was merged there are a few resulting conflicts here - would you mind resolving them?

jeffdgr8:
@barbeau I've merged the latest changes from master and also fixed the code to avoid `NullPointerException`s from tests with a null map.

barbeau:
Thanks @jeffdgr8!

barbeau:
I've created a new multi-layer demo activity in PR https://github.com/googlemaps/android-maps-utils/pull/598 (which is based on this PR branch) that adds the following to a single map to test multiple layer interactions:
 * a cluster manager with markers
 * a GeoJSON polyline layer
 * 2 KML layers (polyline and polygon)
 * a normal marker (unclustered)

I've started testing this PR with the multi-layer demo, and here's what I see so far:
* When all layers are enabled, the click listeners don't seem to fire on the GeoJSON or KML layers
* If I comment out the KML polygon, then the KML polyline listener fires (and the cluster and normal marker respond to clicks) but not the GeoJSON layer
* If I comment out both KML layers, then the remaining listeners work fine (GeoJSON, the cluster, and normal markers)

(Note that KML Polygon click listening is also broken in the master branch - see https://github.com/googlemaps/android-maps-utils/issues/558)

I haven't started digging into why the other items aren't functioning as intended (assuming I understand the intent of this PR correctly), but @jeffdgr8 let me know if you have any thoughts.

jeffdgr8:
@barbeau the object managers work by acting as a hub for the map event listeners, which delegate the specific events to the collections created by the managers in the different map layers. It's required to create these object managers and share them between the layers using the constructors that take them as parameters. I created a PR for your demo https://github.com/CUTR-at-USF/android-maps-utils/pull/1 that does this and confirmed all the click listeners are working, except for the polygon, which is apparently broken (https://github.com/googlemaps/android-maps-utils/issues/558).

I also hadn't added these shared object managers to the GeoJsonLayer yet, so I did this in this commit https://github.com/googlemaps/android-maps-utils/pull/380/commits/609008decd6cebe453f9986adc20f75bfbcc0e1b.

barbeau:
@jeffdgr8 Thanks!! I'll take another look.

barbeau:
@googlebot

jeffdgr8:
@barbeau I made those changes to the doc comments. I also removed the `throws IOException, XmlPullParserException` from `addKMLToMap()`, as these exceptions aren't actually thrown here, only in the `KmlLayer` constructors.

barbeau:
Thanks @jeffdgr8!!

bartekpacia:
That's great! 
Will the official docs be updated with these changes?

barbeau:
@bartekpacia you can find the multilayer docs at https://developers.google.com/maps/documentation/android-sdk/utility/multilayer

PR #381, opened 2017-03-28
Retain the bitmap cache

jeffdgr8:
Use a retained fragment to retain the Renderer bitmap cache between configuration changes.


jeffdgr8:
These changes are important to avoid redownloading all bitmap icons after each configuration change.

library/src/main/java/com/google/maps/android/data/Renderer.java
```diff
@@ -16,8 +16,11 @@
 
 package com.google.maps.android.data;
 
-import android.content.Context;
+import android.app.Activity;
```
barbeau: Could you please use the Android X libraries here instead of the older `android.app` references?

demo/src/main/java/com/google/maps/android/utils/demo/KmlDemoActivity.java
```diff
@@ -42,7 +42,7 @@ public void startDemo () {
 
     private void retrieveFileFromResource() {
         try {
-            KmlLayer kmlLayer = new KmlLayer(mMap, R.raw.campus, getApplicationContext());
+            KmlLayer kmlLayer = new KmlLayer(mMap, R.raw.campus, this);
```
barbeau: Note that this is a breaking change for apps, as any apps passing `getApplicationContext()` would need to change to passing an Activity, as shown here in the demo app. 

@jpoehnelt Normally this would require a new major version release in traditional semantic versioning, but given that we're still pre v1.0 I don't know your feelings on this.

library/src/main/java/com/google/maps/android/data/Renderer.java
```diff
@@ -140,6 +149,29 @@ public Renderer(GoogleMap map, HashMap<? extends Feature, Object> features) {
         mContainerFeatures = null;
     }
 
+    /**
+     * Fragment for retaining the bitmap cache between configuration changes.
+     */
+    public static class RetainFragment extends Fragment {
+        private static final String TAG = RetainFragment.class.getName();
```
barbeau: So this results in a cache in this fragment which is used globally for all activities (i.e., single cache per process). This seems reasonable to me, because if it were scoped per activity you'd likely have duplicates across the caches.

library/src/main/java/com/google/maps/android/data/Renderer.java
jeffdgr8: Done

jeffdgr8:
Thanks for getting to this review. I opted to handle the case of a null activity argument by instantiating the `LruCache` without a retained fragment in this case. Getting an activity for that test case would require adding a dependency, test rule, and test activity, which was easier to avoid by just handling the null.

PR #385, opened 2017-04-08
Storing Collection objects in BiMultiMap

bewhari:
**Changes**
- Explicitly call the Collection version of `put(key, value)` to correctly store the reverse relationship

The method `Renderer#addFeature(Feature feature)` calls the method `BiMultiMap#put` with an `Object` parameter. Java dynamic method dispatch does not work when passing objects as parameters, so the generic method `BiMultiMap#put(K key, Object object)` is always called, even when the created `mapObject` object is of type `Collection`.

### DIFF
--- a/library/src/com/google/maps/android/data/geojson/BiMultiMap.java
+++ b/library/src/com/google/maps/android/data/geojson/BiMultiMap.java
@@ -30,8 +30,12 @@ public class BiMultiMap<K> extends HashMap<K, Object> {
     @Override
     public Object put(K key, Object value) {
         // Store value/key in the reverse map.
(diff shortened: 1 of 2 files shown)


bewhari:
I signed it!

stephenmcd:
Thanks a lot. Is there a test we can add that demonstrates what this fixes?

bewhari:
Thanks for getting back so quickly. I'll add a test in the next commit.

stephenmcd:
Awesome, thanks a lot.

bewhari:
Added a unit test `testCollection` and verified failing on master

stephenmcd:
Great stuff, thanks again!

PR #418, opened 2017-08-30
KmlPlacemark method call over mInlineStyle when it is null

cpuente:
Check for method call over mInlineStyle when it is null


cpuente:
I signed it!

2017-08-30 17:30 GMT+02:00 googlebot <notifications@github.com>:

stephenmcd:
Thanks again

PR #419, opened 2017-08-30
Method to get StyleId's from mStyleMap KmlContainer

cpuente:
Because KmlPlacemarks returns StyleId's that are inside mStyleMap but containers only returns styles inside mStyles it is impossible to get all Styles. 

Add a method (getStyleIdFromMap) to get true style id from map.


stephenmcd:
Thanks

PR #445, opened 2018-03-20
 Added overridden methods to ClusterItem

khairi96:
I added some existing methods to the interface ClusterItem and I made them overridden in MyItem.


khairi96:
I signed it!

khairi96:
I signed it!

stephenmcd:
I don't think there's a strong case for requiring ClusterItem implementors to define the setters you've added, particularly given that this would break existing code-bases. Please let me know if I've missed anything.

khairi96:
Thank you for your consideration. But, I think that using an interface and implementing its methods requires us to add the setters in the interface ClusterItem. These additions, I think are more suitable and they won't affect any other component.

PR #454, opened 2018-04-26
Set polygon clickable option with style instead of forced true.

Qwatcha:
setOnMapClickListener not called when clicking on map area with GeoJson polygon added. Polygon clickable option was set to true. Allow the user to set the clickable option.

Patch file with modified GeoJsonDemoActivity.java to test setting clickable style
[Demo_of_clickable_polygon_or_clicking_through.zip](https://github.com/googlemaps/android-maps-utils/files/1951404/Demo_of_clickable_polygon_or_clicking_through.zip)


stephenmcd:
Thanks

Issue #388, opened 2017-04-21
SphericalUtil.interpolate bug


hptran:
**Summary:** 

SphericalUtil.interpolate(from, to, fraction) does not work when *from* and *to* are close together (about 2 meters)

**Steps to reproduce:** 

select any two locations that are close together and perform interpolation

**Expected behavior:** 

interpolate between two locations in an indoor map

**Observed behavior:** 

convert to XY locations to do interpolation

**Device and Android version:** 

nexus 6

**Screenshots:** 

no screenshot

barbeau:
@hptran could you give us some actual numbers for the input you're using, the expected output, and the actual output? This will help us debug and write unit tests.

jcfarsight:
@barbeau I've encountered the same issue with the following figures and code.

```
Log.d(TAG, "ShortestLine: " + shortestLine.first + ", " + shortestLine.second);
Log.d(TAG, "SecondShortestLine: " + secondShortestLine.first + ", " + secondShortestLine.second);

LatLng pointOne = SphericalUtil.interpolate(shortestLine.first, shortestLine.second, 0.5);
LatLng pointTwo = SphericalUtil.interpolate(secondShortestLine.first, secondShortestLine.second, 0.5);

Log.d(TAG, "PointOne: " + pointOne);
Log.d(TAG, "PointTwo: " + pointTwo);
```

Logs:

> ShortestLine: lat/lng: (-37.756891,175.325262), lat/lng: (-37.756853,175.325242)
> SecondShortestLine: lat/lng: (-37.756831,175.326874), lat/lng: (-37.756764,175.326839)
> PointOne: lat/lng: (-37.756891,175.325262)
> PointTwo: lat/lng: (-37.7567975000013,175.32685649999212)

The interpolation on the second shortest line (length ~8.5m) is working correctly and outputting PointTwo with a believable value. This is not the case however with the shortest line (length ~3.9m) and the output of PointOne appears to be clamped to one of the two input points regardless of the 0.5 fraction passed in.

In this case I would expect a value closer to say (-37.756872, 175.325250) for PointOne.

**Google Play services version:** 42
**Android device:** Nexus 5X
**Android version:** 7.1.2

barbeau:
@jcfarsight Thanks for the numbers!  From a quick look, it appears that the clamping is intentional - if the sine of the angle between the two points is less than `1E-6` then it returns the end point:

[...]

PR #458, opened 2018-07-07
Fall back to linear interpolation when interpolating in short distances (#388)

eggpi:
Hi, I got affected by issue #388 and thought I'd give the linear interpolation solution a try, since it is suggested in the discussion there (and the Wikipedia article linked in the code), and also seemed to give sensible results after trying it out for a couple of examples. Does this seem like a reasonable approach?


stephenmcd:
Great stuff, thank you.

PR #476, opened 2018-09-27
Prevent throw "IllegalArgumentException: left == right"

ipcjs:
see: https://issuetracker.google.com/issues/35827242


ipcjs:
I split them into two PRs：#478 #479

PR #479, opened 2018-10-02
Config for JitPack.io

ipcjs:
Make the project can be automatically compiled on [jitpack.io](https://jitpack.io/#googlemaps/android-maps-utils) by judging whether there is [a JITPACK variable](https://jitpack.io/docs/BUILDING/#build-environment).

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -1,9 +1,4 @@
 apply plugin: 'com.android.library'
-apply plugin: 'maven'
-apply plugin: 'signing'
-
-archivesBaseName = 'android-maps-utils'
-group = 'com.google.maps.android'
 
 dependencies {
     compile 'com.google.android.gms:play-services-maps:11.0.4'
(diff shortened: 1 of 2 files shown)


stephenmcd:
@domesticmouse any thoughts?

stephenmcd:
@ipcjs can you describe what this achieves?

library/build.gradle
```diff
@@ -35,106 +30,15 @@ android {
...
-
-        developers {
-            developer {
-                id 'broady'
-                name 'Chris Broadfoot'
-                url 'http://google.com/+ChristopherBroadfoot'
-            }
-        }
-    }
+if (System.getenv("JITPACK")) {
```
ipcjs: see: https://jitpack.io/docs/BUILDING/#build-environment

ipcjs:
@stephenmcd I updated this description.

barbeau:
My understanding is that JitPack serves as a "maven central" for projects compiled directly from GitHub commits/releases.

@ipcjs in this projects case, I think the main benefit of enabling JitPack is that it would allow users of the library to use any branch or commit as their dependency. So users could use the current master branch head commit as a direct dependency for their project from JitPack, without having to depend on a release or build the project locally themselves. Is this right? Are there other benefits?

ipcjs:
@barbeau Yes, you are right, that is all.

stephenmcd:
Ok that seems very worthwhile - thanks all for the input.

PR #481, opened 2018-11-02
Fix deprecated API usage warnings

jeffdgr8:
Since ScreenBasedAlgorithm's GoogleMap.OnCameraChangeListener onCameraChange() method is only used within ClusterManager, it doesn't need to extend the deprecated listener.


amuramoto:
Thanks for the PR!

Issue #488, opened 2018-11-21
GeoJsonFeature fromJson viaolates same JSON name mId


P72B:
**Summary:** 

I as a user of the android-maps-utils whant to create a GeoJsonFeature from given JSON string. For example new Gson().fromJson(stringifyedGeoJsonFeature, GeoJsonFeature.class).

**Steps to reproduce:** 

Update the android-maps-utils version from 0.4.4 to 0.5+. Parse a valid given JSON string like: new Gson().fromJson(stringifyedGeoJsonFeature, GeoJsonFeature.class). It will crash.

**Expected behavior:** 

The parsing of the JSON string will succeed and I can use the GeoJsonFeature object in my code.

**Observed behavior:** 

This error message will appear:
java.lang.IllegalArgumentException: class com.google.maps.android.data.geojson.GeoJsonFeature declares multiple JSON fields named mId

Background:
Since v0.5+ the class GeoJsonFeature now extends Feature. And both classes have the private member variable mId. My proposal make the mId inside Feature.java protected and the described issue will be gone. Or rename the member mId inside the GeoJsonFeature.

amuramoto:
merged fix from https://github.com/googlemaps/android-maps-utils/pull/490

PR #489, opened 2018-11-21
#488 Made Feature member id protected

P72B:
To prevent same name declaration issues for JSON fields. See issue #488.


P72B:
I signed it! (CLA)

PR #490, opened 2018-11-21
#488 Made Feature member id protected

P72B:
To prevent same name declaration issues for JSON fields. See issue #488.


amuramoto:
Thanks for the PR!

PR #506, opened 2019-03-24
Demo clustering with ViewModel

jeffdgr8:
Move algorithm lock from `ClusterManager` to `Algorithm` to tie the locking to the algorithm object, to better allow reusing the algorithm in multiple `ClusterManager` instances, to best support retaining the algorithm in a ViewModel.

Add a demo of clustering using a view model to retain the cluster item data between configuration changes.

Add `removeItems()` for optimally removing many items, without repeatedly locking and unlocking the algorithm in a tight loop.

~Update all demos to only move map camera on initial load, allowing configuration changes to restore the camera position.~ Moved to https://github.com/googlemaps/android-maps-utils/pull/567

Update URLs to HTTPS (one no longer works with HTTP).


jpoehnelt:
@jeffdgr8 Can you split this up into a couple separate prs?

jeffdgr8:
> @jeffdgr8 Can you split this up into a couple separate prs?

I made a PR with the demo map restore changes by themselves. https://github.com/googlemaps/android-maps-utils/pull/567 After this PR is merged, it'll simplify this diff. I couldn't pull those changes out completely, as there are dependencies in this PR.

Issue #463, opened 2018-07-31
KML test files are bundled in library release


petrdvorak:
```
Android Studio 3.1.3
Build #AI-173.4819257, built on June 4, 2018
JRE: 1.8.0_152-release-1024-b01 x86_64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
Mac OS X 10.13.6
```

**Summary:** 

I am very probably missing something (and I apologize for my ignorance in advance). However, I didn't find any guidelines on this in the [Setup Guide](https://developers.google.com/maps/documentation/android-sdk/utility/setup).

I created a new vanilla app in a fairly new Android Studio with the minimum and target SDK set on 27. Then, I added the following two dependencies to my module `gradle.properties` file:

```rb
implementation 'com.google.android.gms:play-services-maps:15.0.1'
implementation 'com.google.maps.android:android-maps-utils:0.5'
```

After I successfully built the app, I can see the following files in the resulting `*.apk` file:

```
amu_ballon_gx_prefix.kml
amu_basic_folder.kml
amu_basic_placemark.kml
amu_cdata.kml
amu_default_balloon.kml
amu_document_nest.kml
amu_draw_order_ground_overlay.kml
amu_extended_data.kml
amu_ground_overlay_color.kml
amu_ground_overlay.kml
amu_inline_style.kml
amu_multigeometry_placemarks.kml
amu_multiple_placemarks.kml
amu_nested_folders.kml
amu_nested_multigeometry.kml
amu_poly_style_boolean_alpha.kml
amu_poly_style_boolean_numeric.kml
amu_unknwown_folder.kml
amu_unsupported.kml
amu_visibility_ground_overlay.kml
```

Could the documentation mention a way to exclude these unnecessary resources? For example, while I see the benefit of using the Android Maps Utils in our banking app in general, I do not see any benefit of having an `amu_ballon_gx_prefix.kml` file packaged in the app as well, with the following example content:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<kml xmlns="http://www.opengis.net/kml/2.2"
 xmlns:gx="http://www.google.com/kml/ext/2.2">

  <Document>
    <name>balloonVisibility Example</name>
    <open>1</open>
      <Style id="linestyleExample">
        <LineStyle>
          <color>7f0000ff</color>
          <width>4</width>
          <gx:labelVisibility>1</gx:labelVisibility>
        </LineStyle>
      </Style>
    <gx:Tour>
[...]

Issue #371, opened 2017-03-03
Emulator-based unit tests no longer work on Travis


barbeau:
**Summary:** 

There are issues with API Level 23-25 emulators that prevent them from running on Travis.

Here's an example of a failed build:
https://travis-ci.org/googlemaps/android-maps-utils/builds/206977302

Things go unresponsive with the recent master branch (as of https://github.com/googlemaps/android-maps-utils/commit/9be09a2e4a77a222d62d5719bcaa6413dcf373f6).  I've tried a number of resolutions in these branches/PRs (for this and another project):
* https://github.com/barbeau/android-maps-utils/tree/test-travis
* https://github.com/OneBusAway/onebusaway-android/pull/721

...but so far no luck.

It seems to boil down to these two problems with the emulators/SDK:
* API Level 23 - https://code.google.com/p/android/issues/detail?id=228113
* API Level 24/25 - https://code.google.com/p/android/issues/detail?id=230083

@stephenmcd or @broady Any chance you can try to get the Android emulator/SDK team to take a look at the above?  I'm not very familiar with how this is set up, but it may be a simple issue with the metadata associated with the SDK when fetched by something like Travis.

For now, as a workaround in PR https://github.com/googlemaps/android-maps-utils/pull/370, I've changed it to simply build the project, without spinning up an emulator and running the unit tests.  This is obviously a huge inconvenience, as you still need to run the unit tests locally to verify that they pass.

**Steps to reproduce:** 

Run a Travis build as of https://github.com/googlemaps/android-maps-utils/commit/9be09a2e4a77a222d62d5719bcaa6413dcf373f6 in the master branch, with the `.travis.yml`  config shown in https://github.com/googlemaps/android-maps-utils/blob/8f7d4b153d7281c818853474361914d3aefc0678/.travis.yml.

**Expected behavior:** 

Emulator image should be downloaded, started, and the unit tests should run via `./gradlew clean check connectedCheck -x library:signArchives -PdisablePreDex`.

**Observed behavior:** 

The emulator times out (or there is other behavior with missing ABIs, etc. depending on the API Level) - see https://travis-ci.org/googlemaps/android-maps-utils/builds/206977302 for an example, which shows:

~~~
5.61s$ echo no | android create avd --force -n test -t "android-"$ANDROID_API_LEVEL --abi $ANDROID_ABI --tag $ANDROID_TAG
Android 7.0 is a basic Android platform.
[...]

PR #529, opened 2019-06-13
Move test resources to test directory.

ming13:
Addresses #463.


barbeau:
@ming13 Thanks for the PR! 

IIRC these files are used in unit tests so simply deleting them will break the tests. I believe the best fix for this is to move them into the test directory as mentioned in https://github.com/googlemaps/android-maps-utils/issues/463#issuecomment-410360845.

ming13:
@barbeau, yeah, I read about that but neither code search or Save Delete option in IJ have found any usage. Renamed the PR and adjusted the change.

BTW, does anybody run tests? I had to fix a bunch of them to be green. And CI is not running them (and running at all as it seems). This is scary.

library/build.gradle
```diff
@@ -34,6 +35,9 @@ android {
         installOptions "-d", "-t"
     }
 
+    defaultConfig {
+        testInstrumentationRunner 'android.support.test.runner.AndroidJUnitRunner'
+    }
```
ming13: Tests were not found without the runner. No idea how it worked before (if it did).

library/tests/src/com/google/maps/android/clustering/StaticClusterTest.java
```diff
@@ -19,13 +19,15 @@
...
 
+import org.junit.Before;
 import org.junit.Test;
 import org.junit.Assert;
 
 public class StaticClusterTest {
 
     private StaticCluster<ClusterItem> mCluster;
 
+    @Before
```
ming13: `setUp` was not called without the annotation.

library/tests/src/com/google/maps/android/data/LineStringTest.java
```diff
@@ -48,19 +48,19 @@ public void testGetGeometryObject() throws Exception {
...
-        Assert.assertEquals(lineString.getGeometryObject().get(0).latitude, 90.0);
-        Assert.assertEquals(lineString.getGeometryObject().get(1).latitude, 90.0);
-        Assert.assertEquals(lineString.getGeometryObject().get(2).latitude, 90.0);
-        Assert.assertEquals(lineString.getGeometryObject().get(3).longitude, 53.0);
-        Assert.assertEquals(lineString.getGeometryObject().get(4).longitude, 54.0);
+        Assert.assertEquals(lineString.getGeometryObject().get(0).latitude, 90.0, 0);
+        Assert.assertEquals(lineString.getGeometryObject().get(1).latitude, 90.0, 0);
+        Assert.assertEquals(lineString.getGeometryObject().get(2).latitude, 90.0, 0);
+        Assert.assertEquals(lineString.getGeometryObject().get(3).longitude, 53.0, 0);
+        Assert.assertEquals(lineString.getGeometryObject().get(4).longitude, 54.0, 0);
```
ming13: This and a bunch of similar changes address this:

```
java.lang.AssertionError: Use assertEquals(expected, actual, delta) to compare floating-point numbers
    at org.junit.Assert.fail(Assert.java:88)
    at org.junit.Assert.assertEquals(Assert.java:667)
    at org.junit.Assert.assertEquals(Assert.java:656)
```

barbeau:
CI issue is discussed in https://github.com/googlemaps/android-maps-utils/issues/371.

Yes, the project probably needs some updating to run tests correctly in Android Studio.  Let's split those changes out to their own PRs though so each serves it's own purpose.

Looking at my old comment the KML files are used in `library/tests/src/com/google/maps/android/kml/Kml*Test.java` classes - I don't believe Android Studio will flag them via a Find on the files.

ming13:
@barbeau, maybe it’ll be fine to merge moving files and fixing tests together? I mean, it literally makes tests runnable, it wouldn’t be even possible to merge this PR if the CI would be configured properly to run tests.

ming13:
@jpoehnelt, @friederbluemle, please take a look. We had open questions so I’m just waiting for the clarification.

jpoehnelt:
I've only just started taking a look at this repo(we may have some more resources to help out soon). I don't see anything of note from a quick glance but I would like to run locally before approving. I'll try to get to it by mid next week. Thanks again!

library/build.gradle
```diff
@@ -34,6 +35,9 @@ android {
         installOptions "-d", "-t"
     }
 
+    defaultConfig {
+        testInstrumentationRunner 'android.support.test.runner.AndroidJUnitRunner'
```
friederbluemle: There already is a `defaultConfig` section in this file. Can you merge it with the existing one?

friederbluemle:
@jpoehnelt - Merging the base branch into the PR head branch won't help here - There are a few code changes required to properly fix the test setup on the latest `master` branch. Should I open a PR?

ming13:
Hm, seems like Travis passed, but seems like it doesn’t run tests for now.

ming13:
@friederbluemle, since you’ve resolved test issues separately — I think it would be better to merge your changes first and I’ll rebase mine on top. Also — can I ask to add the `test` Gradle task in Travis configuration on your branch? Don’t want to break anything and it is better to have CI check that.

friederbluemle:
@ming13 - Sure, I'll open a PR with just the test setup fixes. 👍 
I definitely agree this is something that should run on CI with every PR. There are more test updates and fixes on the way, hopefully we'll get there soon. Right now, running the `test` task wouldn't do much more than the existing `check` task does. The tests are setup as "instrumented" tests, and require a connected Android device (something that unfortunately is not that straight forward on Travis CI).

ming13:
@friederbluemle, `testDebugUnitTest` then? I think that tests both of us fixed can be run on the JVM.

friederbluemle:
@ming13 - Yes, you are right, it _should_ be possible to run the tests on the JVM only. There are no real "UI" tests involved. However, since the code tested does contain imports from the Android framework, it may require non trivial additional changes to the test setup, so it might be better to have a separate PR for that.

friederbluemle:
Good news @ming13 - Your branch rebases cleanly onto the latest `master` (with #550 merged). All you need to do is remove the `defaultConfig` block from build.gradle. :)

library/build.gradle
ming13: Removed.

ming13:
@jpoehnelt, @friederbluemle, rebased and squashed! Tests passed locally: 184 of 184.

friederbluemle:
Awesome, thanks @ming13 👏

PR #550, opened 2019-10-03
Fix test setup

friederbluemle:
Background: @ming13 initially opened #529 on June 13 which contained similar changes. Back then, the project was not using androidx dependencies yet, so a couple more changes were required to make it work with the latest `master` branch.

To test this, run:

```
./gradlew :library:connectedCheck
```

Notice that on the current `master` branch, the tests are broken. With this PR merged, the existing tests will pass.

More follow-up test fixes will be required - The rebased #529 will move test resources, and hopefully we can convert these to pure unit tests so that a connected device is no longer required.

For now, these are the _minimal_ changes required to restore the ability to run the existing tests.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -8,8 +8,8 @@ dependencies {
 
     testImplementation 'junit:junit:4.12'
     androidTestImplementation 'junit:junit:4.12'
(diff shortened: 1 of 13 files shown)


friederbluemle:
Thanks for the quick merge @jpoehnelt 👍

PR #599, opened 2020-01-09
Fix #427 - Trim leading/trailing whitespace in KML colors

### LINKED ISSUE #427: kml: Whitespace in <color> throws exception
Extra whitespace in a <color> tag will throw a NumberFormatException from Color.parseColor

For example:

         <color>ff0098e6 </color> 

It's not immediately obvious from the XML and KML specs whether this is valid, but in any case, it would make sense to trim the color string before attempting to parse it.

A sample KML with this extra whitespace is https://rmgsc.cr.usgs.gov/outgoing/GeoMAC/ActiveFirePerimeters.kml

Here is a stack trace:

Caused by: java.lang.NumberFormatException: Invalid long: "ff0098e6 "
                                                                                 at java.lang.Long.invalidLong(Long.java:124)
                                                                                 at java.lang.Long.parse(Long.java:361)
                                                                                 at java.lang.Long.parseLong(Long.java:352)
                                                                                 at android.graphics.Color.parseColor(Color.java:211)
                                                                                 at com.google.maps.android.data.kml.KmlStyle.setOutlineColor(KmlStyle.java:334)
                                                                                 at com.google.maps.android.data.kml.KmlStyleParser.createLineStyle(KmlStyleParser.java:195)
                                                                                 at com.google.maps.android.data.kml.KmlStyleParser.createStyle(KmlStyleParser.java:55)
                                                                                 at com.google.maps.android.data.kml.KmlContainerParser.setContainerStyle(KmlContainerParser.java:148)
                                                                                 at com.google.maps.android.data.kml.KmlContainerParser.assignPropertiesToContainer(KmlContainerParser.java:91)
                                                                                 at com.google.maps.android.data.kml.KmlContainerParser.createContainer(KmlContainerParser.java:52)
                                                                                 at com.google.maps.android.data.kml.KmlParser.parseKml(KmlParser.java:74)
                                                                                 at com.google.maps.android.data.kml.KmlLayer.<init>(KmlLayer.java:47)

barbeau:
This PR avoids exceptions when parsing KML files that include leading or trailing spaces in colors, like `<color>ff0098e6 </color>`, by trimming leading and trailing whitespace prior to parsing the color. This applies to colors for markers, polylines, and polygons. Unit tests are included.

It turns out KML colors with leading whitespace were already tolerated in the library, but was treated as a special case and assumed that the leading whitespace was actually intended to be a `0`. This behavior in the library was apparently due to previous behavior of Google Map's KML export feature. From `KmlStyle.convertColor()`:

~~~
        // Maps exports KML colors with a leading 0 as a space.
        if (newColor.substring(0, 1).equals(" ")) {
            newColor = "0" + newColor.substring(1, newColor.length());
        }
~~~

I've checked Google Maps KML export and it no longer seems to have this issue, so I've removed this code and the corresponding unit test.

So, we now effectively support the exact KML color spec `aabbggrr`:
https://developers.google.com/kml/documentation/kmlreference#-colorstyle-

...just with tolerance for leading or trailing whitespace.

### DIFF
--- a/library/src/androidTest/java/com/google/maps/android/data/kml/KmlStyleTest.java
+++ b/library/src/androidTest/java/com/google/maps/android/data/kml/KmlStyleTest.java
@@ -1,7 +1,9 @@
 package com.google.maps.android.data.kml;
 
 import com.google.android.gms.maps.MapsInitializer;
+import com.google.maps.android.TestUtil;
 
+import org.junit.Assume;
 import org.junit.Before;
 import org.junit.Test;
 
@@ -113,6 +115,10 @@ public class KmlStyleTest {
 
     @Test
     public void testMarkerColorLeadingSpace() {
+        if (TestUtil.isRunningOnTravis()) {
+            Assume.assumeTrue("Skipping KmlStyleTest.testMarkerColorLeadingSpace() - this is expected behavior on Travis CI (#573)", false);
+            return;
+        }
         KmlStyle kmlStyle = new KmlStyle();
         assertNotNull(kmlStyle);
         assertNotNull(kmlStyle.getMarkerOptions());
@@ -122,6 +128,10 @@ public class KmlStyleTest {
 
     @Test
     public void testMarkerColorTrailingSpace() {
+        if (TestUtil.isRunningOnTravis()) {
+            Assume.assumeTrue("Skipping KmlStyleTest.testMarkerColorTrailingSpace() - this is expected behavior on Travis CI (#573)", false);
+            return;
+        }
         KmlStyle kmlStyle = new KmlStyle();
         assertNotNull(kmlStyle);
         assertNotNull(kmlStyle.getMarkerOptions());
--- a/library/src/androidTest/java/com/google/maps/android/data/kml/KmlStyleTest.java
+++ b/library/src/androidTest/java/com/google/maps/android/data/kml/KmlStyleTest.java
@@ -1,12 +1,26 @@
 package com.google.maps.android.data.kml;
 
-import android.graphics.Color;
+import com.google.android.gms.maps.MapsInitializer;
 
+import org.junit.Before;
 import org.junit.Test;
 
-import static org.junit.Assert.*;
+import android.graphics.Color;
+
+import androidx.test.platform.app.InstrumentationRegistry;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 
 public class KmlStyleTest {
+
+    @Before
+    public void setUp() {
+        MapsInitializer.initialize(InstrumentationRegistry.getInstrumentation().getTargetContext());
+    }
+
     @Test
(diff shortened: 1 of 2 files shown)


barbeau:
The first Travis build of this PR failed because `MapsInitializer.initialize(...)` apparently still isn't working on the emulators. I just pushed another commit disabling these two tests that require it when running on Travis, so hopefully this time it works :crossed_fingers:.

library/src/androidTest/java/com/google/maps/android/data/kml/KmlStyleTest.java
```diff
@@ -81,9 +114,54 @@ public void testLineColor() {
...
+    public void testMarkerColorLeadingSpace() {
+        if (TestUtil.isRunningOnTravis()) {
+            Assume.assumeTrue("Skipping KmlStyleTest.testMarkerColorLeadingSpace() - this is expected behavior on Travis CI (#573)", false);
+            return;
+        }
         KmlStyle kmlStyle = new KmlStyle();
         assertNotNull(kmlStyle);
         assertNotNull(kmlStyle.getMarkerOptions());
+        kmlStyle.setMarkerColor(" FFFFFF");
+        // We can't get the marker color to see if it set correctly
```
arriolac: How about updating the visibility of [KmlStyle.mMarkerColor](https://github.com/googlemaps/android-maps-utils/blob/master/library/src/main/java/com/google/maps/android/data/kml/KmlStyle.java#L46) and annotate it with [VisibleForTesting](https://developer.android.com/reference/androidx/annotation/VisibleForTesting)? Otherwise, without inspecting the resulting marker color this test and `testMarkerColorTrailingSpace` should be removed.

library/src/main/java/com/google/maps/android/data/kml/KmlStyle.java
```diff
@@ -187,6 +187,8 @@ public String getIconUrl() {
      * @param color Fill color for a KML Polygon as a String
      */
     /* package */ void setFillColor(String color) {
+        // Tolerate KML with leading or trailing whitespace in colors
+        color = color.trim();
```
arriolac: How about applying this method inside `convertColor` to avoid duplication and to make sure the same input sanitation is applied? Looks like in all cases `color.trim()` is desired before calling `convertColor` anyway.

barbeau: Makes sense! I just refactored this accordingly in https://github.com/googlemaps/android-maps-utils/pull/599/commits/5074520c342ef46fc6eb82f002918609a4f0c350.

library/src/androidTest/java/com/google/maps/android/data/kml/KmlStyleTest.java
barbeau: Sounds good - I just changed this in https://github.com/googlemaps/android-maps-utils/pull/599/commits/82f131d93a523fdd3f252c96dc5903a30fc33515. FWIW the existing unit test was still ensuring that an exception wasn't thrown, but I agree this is a better solution.

PR #600, opened 2020-01-09
Fix #530 - Check for empty KML Style <hotspot> element before parsing

### LINKED ISSUE #530: Crash when loading KML File if hotspot tag is empty
**Summary:** 

When i load  kml file via raw resource in which "hotspot" tag is an empty app crash
**Steps to reproduce:** 

simply load 20 MB kml file via following code:

`var layer = KmlLayer(mMap, R.raw.campus, context)`

**Expected behavior:** 

object of layer should return
load kml to map

**Observed behavior:** 

crash

**Device and Android version:** 

Android 9
Redmi Note 7 PRO

**Crash Exception :** 

~~~
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String java.lang.String.trim()' on a null object reference
        at sun.misc.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:1838)
        at sun.misc.FloatingDecimal.parseFloat(FloatingDecimal.java:122)
        at java.lang.Float.parseFloat(Float.java:451)
        at com.google.maps.android.data.kml.KmlStyleParser.setIconHotSpot(KmlStyleParser.java:176)
        at com.google.maps.android.data.kml.KmlStyleParser.createIconStyle(KmlStyleParser.java:95)
        at com.google.maps.android.data.kml.KmlStyleParser.createStyle(KmlStyleParser.java:53)
        at com.google.maps.android.data.kml.KmlContainerParser.setContainerStyle(KmlContainerParser.java:148)
        at com.google.maps.android.data.kml.KmlContainerParser.assignPropertiesToContainer(KmlContainerParser.java:91)
        at com.google.maps.android.data.kml.KmlContainerParser.createContainer(KmlContainerParser.java:52)
        at com.google.maps.android.data.kml.KmlParser.parseKml(KmlParser.java:74)
        at com.google.maps.android.data.kml.KmlLayer.<init>(KmlLayer.java:47)
        at com.google.maps.android.data.kml.KmlLayer.<init>(KmlLayer.java:29)
~~~

barbeau:
This prevents a crash due to NPE when parsing `<hotSpot />` within KML styles.  Test file snippet and new unit test are included.

### DIFF
--- a/library/src/main/java/com/google/maps/android/data/kml/KmlStyleParser.java
+++ b/library/src/main/java/com/google/maps/android/data/kml/KmlStyleParser.java
@@ -3,8 +3,6 @@ package com.google.maps.android.data.kml;
 import org.xmlpull.v1.XmlPullParser;
 import org.xmlpull.v1.XmlPullParserException;
 
-import android.text.TextUtils;
-
 import java.io.IOException;
 import java.util.HashMap;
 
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/google/maps/android/data/kml/KmlStyleParser.java
```diff
@@ -172,6 +174,10 @@ private static void setIconUrl(XmlPullParser parser, KmlStyle style)
     private static void setIconHotSpot(XmlPullParser parser, KmlStyle style) {
         float xValue, yValue;
         String xUnits, yUnits;
+        if (TextUtils.isEmpty(parser.getAttributeValue(null, "x")) ||
```
arriolac: It'd be better if an early return that checks if the method [XmlPullParser.isEmptyElementTag](https://developer.android.com/reference/org/xmlpull/v1/XmlPullParser.html#isEmptyElementTag()) is `true` so that the intent is clearer.

barbeau: Thanks, makes sense! I just made the change in https://github.com/googlemaps/android-maps-utils/pull/600/commits/edfcd62244466247968721878a51c30b67990b98.

PR #601, opened 2020-01-22
Use thread pools

jeffdgr8:
Reuse threads from pool instead of creating new thread for every task

### DIFF
--- a/library/src/main/java/com/google/maps/android/clustering/view/DefaultClusterRenderer.java
+++ b/library/src/main/java/com/google/maps/android/clustering/view/DefaultClusterRenderer.java
@@ -64,6 +64,8 @@ import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/google/maps/android/clustering/view/DefaultClusterRenderer.java
```diff
@@ -290,7 +293,7 @@ public void run() {
             });
             renderTask.setProjection(projection);
             renderTask.setMapZoom(mMap.getCameraPosition().zoom);
-            new Thread(renderTask).start();
+            mExecutor.execute(renderTask);
```
barbeau: Just calling out for future reference that given this Executor is instantiated via `newSingleThreadExecutor()` we're changing to single-threading here. But, the entire implementation appears to be effectively single-threaded anyway given the `synchronized` block and `mViewModificationInProgress` variable which will stop any other RenderTasks from being executed until this one finishes. So :+1:

jeffdgr8: That was my line of thinking exactly. There's only ever one render task in flight and potentially one queued up to run next.

PR #606, opened 2020-01-28
feat: Support KML BalloonStyle template

jeffdgr8:
Substitute property values from placemark in BalloonStyle template. See [KML docs](https://developers.google.com/kml/documentation/extendeddata#balloonStyle).


library/src/main/java/com/google/maps/android/data/Renderer.java
```diff
@@ -1028,6 +1030,28 @@ private void setMarkerInfoWindow(KmlStyle style, Marker marker,
...
     }
 
+    /**
+     * Substitute property values in BalloonStyle text template
+     *
+     * @param template text template
+     * @param placemark placemark to get property values from
+     * @return string with property values substituted
+     */
+    private String substituteProperties(String template, KmlPlacemark placemark) {
```
barbeau: Could you please move this to a utility method and add some unit tests to make it more obvious what this is doing?

jeffdgr8: I added a test. Is there a utility class that would make sense to move the method to?

barbeau: Hmmm, doesn't look like there is an existing utility class where this fits.  Maybe a new package scoped class `.data.kml,KmlUtil` ?

I'm trying to avoid stuffing more KML or GeoJSON-specific code in Renderer, which should be more of a generic implementation. And an entire refactor is beyond the scope of this PR.

jeffdgr8: Yeah, that makes sense for sure. I started a `KmlUtil` class with this method.

barbeau:
Thanks @jeffdgr8!

PR #609, opened 2020-01-29
perf: Improve KML bitmap caching and fix icon scaling

jeffdgr8:
Improve KML bitmap caching by caching `BitmapDescriptor`s instead of `Bitmap`s. Every call to `BitmapDescriptorFactory.fromBitmap()` copies the `Bitmap` passed to it. So the current code ends up duplicating every `Bitmap` that is reused for multiple `Marker`s or `GroundOverlay`s for every usage.

Tested with this KML with 25 different marker icons and 25,000 random map points:
[huge_kml.kml](https://github.com/googlemaps/android-maps-utils/files/4126096/huge_kml.kml.txt) (remove .txt extension). [Credit](http://kml4earth.appspot.com/bigKml.html)

On a Pixel 4 XL, the current code uses 0.9 GB RAM.
Screenshot 2020-01-28 at 7 59 36 PM

With these changes, only 294 MB RAM is used.
Screenshot 2020-01-28 at 9 26 02 PM

Note the caches use `HashMap`s and not `LruCache`, as was used previously. This is because only the `BitmapDescriptor`s are cached indefinitely and the cache is only holding a reference to the same `BitmapDescriptor`s used by the `GoogleMap` in its own view. So discarding any of these references wouldn't regain memory, as long as the `BitmapDescriptor`s are still being used by the map view.

The original `Bitmap`s are only retained until all the images have been loaded, scaled, and cached as `BitmapDescriptor`s.

Also fixes scaling for point marker icon images, taking into account screen density and scaling to a [normalized minimum width of 32](https://stackoverflow.com/a/22569590/387576).


library/src/main/java/com/google/maps/android/data/Renderer.java
```diff
@@ -94,9 +96,30 @@
...
+     */
+    private Map<String, BitmapDescriptor> mGroundOverlayImagesCache;
+
+    /**
+     * Map of image URL to Bitmap
+     *
+     * Holds initial references to bitmaps so they can be scaled and BitmapDescriptors cached.
+     * This cache is cleared once all icon URLs are loaded, scaled, and cached as BitmapDescriptors.
+     */
+    private Map<String, Bitmap> mBitmapCache;
```
arriolac: Suggestion: thoughts on wrapping `mMarkerImagesCache`,  `mGroundOverlayImagesCache` and `mBitmapCache` into a class? I think this would help with readability and simplify the API a bit.

library/src/main/java/com/google/maps/android/data/Renderer.java
```diff
@@ -326,9 +363,9 @@ public Feature getContainerFeature(Object mapObject) {
     /**
      * Gets the URLs stored for the Marker icons
      *
-     * @return mMarkerIconUrls ArrayList of URLs
+     * @return mMarkerIconUrls Set of URLs
      */
-    public ArrayList<String> getMarkerIconUrls() {
```
arriolac: Noting that this is another breaking change. I do think this is a break API though.

library/src/main/java/com/google/maps/android/data/Renderer.java
```diff
@@ -94,9 +96,30 @@
...
-    private final ArrayList<String> mMarkerIconUrls;
+    private final Set<String> mMarkerIconUrls;
+
+    /**
+     * Map of image URL to map of scale factor to BitmapDescriptors for point marker icons
+     *
+     * BitmapDescriptors are cached to avoid creating new BitmapDescriptors for each individual
+     * usage of a Bitmap. Each BitmapDescriptor copies the Bitmap it's created from.
+     */
+    private Map<String, Map<Double, BitmapDescriptor>> mMarkerImagesCache;
```
arriolac: The probability of missing the cache can be pretty high having `Double` as a key for `Map`. I suggest storing this as `String` instead and have up to a certain precision level (e.g. 4 decimal places?)

library/src/main/java/com/google/maps/android/data/Renderer.java
```diff
@@ -351,12 +388,77 @@ public Feature getContainerFeature(Object mapObject) {
...
+     * needed for Marker icon images
      *
-     * @return mImagesCache
+     * @param url URL to get cached image for
+     * @param scale scale to get image at
+     * @return scaled BitmapDescriptor
      */
-    public LruCache<String, Bitmap> getImagesCache() {
-        return mImagesCache;
+    public BitmapDescriptor getCachedMarkerImage(String url, double scale) {
```
arriolac: I'd expect that getting a cached image would be pretty fast just from this method name. However, since this method will also perform image scaling, which is time consuming, it would be good to add a note that this is the case.

library/src/main/java/com/google/maps/android/data/Renderer.java
```diff
@@ -351,12 +388,77 @@ public Feature getContainerFeature(Object mapObject) {
...
-     * Gets the images cache which is needed to download GroundOverlays and Marker Icon images
+     * Gets a cached image at the specified scale which is
+     * needed for Marker icon images
      *
-     * @return mImagesCache
+     * @param url URL to get cached image for
+     * @param scale scale to get image at
+     * @return scaled BitmapDescriptor
      */
-    public LruCache<String, Bitmap> getImagesCache() {
```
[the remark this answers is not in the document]
jeffdgr8: Yes, I don't believe anyone would actually be using the `Renderer` API directly. Many of these methods should be protected, as they're used only by the `Renderer` subclasses, `KmlRenderer` and `GeoJsonRenderer`. Many of the public methods can't be package-private or protected only because they're used both in `Layer` (same package) and one of the subclasses, which are in different packages.

Code that utilizes the KML and GeoJSON functionality of the library will mostly be interacting with the `Layer` subclasses, `KmlLayer` and `GeoJsonLayer`, which don't even expose their `Renderer` objects directly in order to access those public methods. `Renderer` could almost be an abstract class if the code was refactored some. There's quite a bit of code that's specific to KML and GeoJSON that could also be moved to one or the other subclasses.

I went ahead and reduced the access scope on all the `Renderer` methods that could be, to either package-private or protected. In this particular method's case, a `Renderer` subclass, which I don't expect to exist outside of `KmlRenderer` and `GeoJsonRenderer`, would follow the pattern in these subclasses and call `getCachedMarkerImage()` or `getCachedGroundOverlayImage()` instead of accessing a cache directly as was previously done through this `getImagesCache()` method.

library/src/main/java/com/google/maps/android/data/Renderer.java
jeffdgr8: As before, I don't believe anyone would actually be using the `Renderer` API directly. I've made this particular method protected now under this assumption. The previous `ArrayList` was always only adding unique icon URLs, but using a set makes this much more efficient, avoiding full list scanning for `contains()` each time one is added.

library/src/main/java/com/google/maps/android/data/Renderer.java
jeffdgr8: Good call. Done: https://github.com/googlemaps/android-maps-utils/pull/609/commits/e9d8f24609e3055861f9961ffcde5f3fe5a09485

library/src/main/java/com/google/maps/android/data/Renderer.java
jeffdgr8: I added more details to the doc comment: https://github.com/googlemaps/android-maps-utils/pull/609/commits/f99a787285ef84fd0b57fc77cfd8e38f045c9055

library/src/main/java/com/google/maps/android/data/Renderer.java
jeffdgr8: I had thought about this briefly and concluded that while a `Double` may not be able to perfectly represent all parsed decimal strings, 0.1 for example, that at least the parsed string double value will be equal for the scale value used to cache the image and scale value used to access the cached scaled image. In other words, a double representation of 0.1 will equal another double representation of 0.1, both obtained from the same parse routine. In fact, the same double `mScale` variable from a `KmlStyle` is often being used to both cache and retrieve the cached images.

However, while I made this conclusion based on the current implementation, I do see how a bug could possibly be introduced in the future, if style scales were introduced from another source somehow, possibly with a lower float precision.

I made this change to use the string formatted scale value: https://github.com/googlemaps/android-maps-utils/pull/609/commits/f9bd8baac9ce104b605e1221edd99d79ba0d7df2

library/src/main/java/com/google/maps/android/data/Renderer.java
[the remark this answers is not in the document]
arriolac: Agreed that scoping should be reduced in that class. Thanks for addressing that!

PR #614, opened 2020-01-30
fix: Make KML/GeoJSON lines and polygons clickable by default

### LINKED ISSUE #558: KML/GeoJSON polygon onFeatureClick() not called
關鍵在於 Renderer 的 addedPolygon.setClickable(polygonOptions.isClickable());

polygonOptions.isClickable() 總是得到false

所以如果想讓 KmlDemoActivity的onFeatureClick 可以被調用

必須修正成 addedPolygon.setClickable(true);

barbeau:
As discussed in https://github.com/googlemaps/android-maps-utils/issues/558, the original intent when [combining the KML and GeoJSON renderers](https://github.com/googlemaps/android-maps-utils/pull/351/files#diff-dea26f53a40d31f061e3537537389db6R683) was to have GeoJSON and KML lines and polygons clickable by default. 

However, there was a regression in https://github.com/googlemaps/android-maps-utils/pull/454 that unintentionally disabled clicks for KML and GeoJSON polygons when attempting to allow developers to use GeoJSON custom styles to set clickable.

I investigated both KML and GeoJSON lines and polygons, and here is the current state of the library:
* For KML/GeoJSON polygons - we're currently supporting developer-defined clickable styles for GeoJSON (the change in https://github.com/googlemaps/android-maps-utils/pull/454), but that results in a default of not clickable for GeoJSON and KML. And there isn't a way to override the default in `KmlLayer`.
* For KML/GeoJSON lines - we're currently ignoring developer-defined clickable styles and hard-coding the default clickable style to true.

This PR solves both problems (default of "not clickable" and supporting developer-defined styles) as follows:
* Makes all styles for GeoJSON and KML lines and polygons clickable by default by setting clickable=true in style constructors rather than the renderer. This also allows the developer to set a custom style with clickable=false on GeoJSON lines and polygons (custom programmatic KML styles currently aren't supported by the library).
* Make clickable default to true in the `Style` superclass - this should hopefully avoid a similar issue in any new implementations that extend `Style`
* Normalize the `GeoJsonLineStringStyle` and `GeoJsonPolygonStyle` clickable implementations - `GeoJsonPolygonStyle` was missing code related to the clickable attribute
* Add missing `<name>` attribute to the KML polygon demo file used in the `MultiLayerDemoActivity`
* Add unit tests:
    * For GeoJSON - test both default and programmatic custom clickable styles for both the style and renderer classes
    * For KML - test the default style for both the style and renderer classes
    * Refactor KML and GeoJSON test utility methods used in more than one test class to new Kml/GeoJsonTestUtil classes
    * Fix GeoJSON polygonStyle.toPolygonOptions().isVisible() test (this previously didn't check PolygonOptions)

I've also added TODO statements in the unit tests where we could expand coverage further (i.e., after adding the layers to the map), but that requires additional test instrumentation to have a live `GoogleMap` object.

Fixes #558

Refs: 558


library/src/androidTest/java/com/google/maps/android/data/kml/KmlTestUtil.java
```diff
@@ -0,0 +1,37 @@
...
+        InputStream stream =
+                InstrumentationRegistry.getInstrumentation()
+                        .getTargetContext()
+                        .getResources()
+                        .openRawResource(res);
+        XmlPullParserFactory factory = XmlPullParserFactory.newInstance();
+        factory.setNamespaceAware(true);
+        XmlPullParser parser = factory.newPullParser();
+        parser.setInput(stream, null);
+        parser.next();
```
arriolac: Why does `next` have to be invoked here?

barbeau: Not sure why it was originally included (this was a direct refactor from another existing test class). Looks like it's not needed though, as I just removed it and the tests all pass without it (and `KmlLayer.createXmlParser()` also doesn't include it). Just pushed https://github.com/googlemaps/android-maps-utils/pull/614/commits/741667f3b53c481348574a76eeed94ce96a79eae with this change, and I'll merge after the CI passes.

Issue #619, opened 2020-02-08
KML load images included in kmz zip file


chitgoks:
Currently images included in the zip/kmz file are not loaded in KMLLayer where the urls in the KML are relative paths to the image.

e.g. /sampel.png

im not sure though if the url starts with http if it also loads, but these kind of urls should also be supported.

barbeau:
@chitgoks Thanks! Could you please provide an example KMZ for this?

chitgoks:
i will PM back once i can get a KMZ for this. i access the google crisis site for the kmz, currently there is no natural disaster.

jeffdgr8:
I added KMZ support in https://github.com/googlemaps/android-maps-utils/pull/625.

Issue #472, opened 2018-09-08
KMZ Support


chitgoks:
Hi are there plans to have KMLLayer support KMZ with images?

There are some KMZ files that contain images which is used by the KML file.

amuramoto:
No plans at this time, but please consider sending a PR to add support if you'd like to see it sooner!

chitgoks:
sorry i dont follow. PR as in pull request? doesn't sound right. and where exactly do i file one?

amuramoto:
Yes, pull request. You can implement support and then send us a pull request for review and potential merge into the codebase.

Some info about PRs and how they work is available here: https://help.github.com/en/articles/about-pull-requests

jeffdgr8:
I added KMZ support in https://github.com/googlemaps/android-maps-utils/pull/625.

PR #625, opened 2020-02-14
feat: KMZ support

jeffdgr8:
Support opening a KMZ file through the same mechanism as opening a KML file, either from a resource ID or `InputStream`, in `KmlLayer`. Images within the KMZ archive are decoded and cached for points and ground overlays the same as downloaded images from URLs.

Fixes https://github.com/googlemaps/android-maps-utils/issues/619, https://github.com/googlemaps/android-maps-utils/issues/472


library/src/main/java/com/google/maps/android/data/Renderer.java
```diff
@@ -553,7 +553,7 @@ private void putMarkerImagesCache(String url, String scale, BitmapDescriptor bit
      * @param url    image URL
      * @param bitmap image bitmap
      */
-    protected void cacheBitmap(String url, Bitmap bitmap) {
+    public void cacheBitmap(String url, Bitmap bitmap) {
```
arriolac: Please revert—changing the scope of `cacheBitmap` and `checkClearBitmapCache` are implementation leaks. I'll comment on `KmlLayer` on another suggested approach.

library/src/main/java/com/google/maps/android/data/kml/KmlLayer.java
```diff
@@ -88,11 +100,13 @@ public KmlLayer(GoogleMap map, int resourceId, FragmentActivity activity, Marker
...
      * Creates a new KmlLayer object - addLayerToMap() must be called to trigger rendering onto a map.
      *
+     * Constructor may be called on a background thread, as I/O and parsing may be long-running.
+     *
      * Use this constructor with shared object managers in order to handle multiple layers with
      * their own event handlers on the map.
      *
      * @param map    GoogleMap object
-     * @param stream InputStream containing KML file
+     * @param stream InputStream containing KML or KMZ file
```
arriolac: I think it would be simpler if another constructor was created, with the following signature, to handle the KMZ files explicitly:

```java
public KmlLayer(GoogleMap map, ZipInputStream stream, FragmentActivity activity, MarkerManager markerManager, PolygonManager polygonManager, PolylineManager polylineManager, GroundOverlayManager groundOverlayManager)
```

This way, it's left to the caller to use the correct constructor for their use-case.

library/src/main/java/com/google/maps/android/data/kml/KmlLayer.java
```diff
@@ -106,14 +120,48 @@ public KmlLayer(GoogleMap map, InputStream stream, FragmentActivity activity, Ma
...
+        ZipInputStream zip = new ZipInputStream(bis);
+        try {
+            KmlParser parser = null;
+            ZipEntry entry = zip.getNextEntry();
+            if (entry != null) { // is a KMZ zip file
+                while (entry != null) {
+                    if (parser == null && entry.getName().toLowerCase().endsWith(".kml")) {
+                        parser = parseKml(zip);
+                    } else {
+                        Bitmap bitmap = BitmapFactory.decodeStream(zip);
```
arriolac: Since other file types might also be used (e.g. `.mp3`), it would be good to check that an image is used and if not, log via `Log.w` that the file type is not supported.

arriolac: Will this handle subdirectories?

```
sample.kmz
 - doc.kml
 - image.png
 - dir/image2.png
 - dir/image3.png
 - dir/dir2/image4.png
 ...
```

library/src/main/java/com/google/maps/android/data/kml/KmlLayer.java
```diff
@@ -106,14 +120,48 @@ public KmlLayer(GoogleMap map, InputStream stream, FragmentActivity activity, Ma
...
+                        if (bitmap != null) {
+                            renderer.cacheBitmap(entry.getName(), bitmap);
```
arriolac: Instead of caching the bitmap in `KmlLayer`, a helper method in `KmlRenderer` would be cleaner:

e.g. 
```java
KmlRenderer.storeKmzData(HashMap<String, KmlStyle> styles,
    HashMap<String, String> styleMaps,
    HashMap<KmlPlacemark, Object> features, ArrayList<KmlContainer> folders,
    HashMap<KmlGroundOverlay, GroundOverlay> groundOverlays,
    Set<String> bitmapInputStreams)
```

This way, all bitmap caching is done within `KmlRenderer` and the scoping of the functions within `Renderer` can be reverted.

library/src/main/java/com/google/maps/android/data/kml/KmlLayer.java
jeffdgr8: I thought about that. Since we can differentiate between the two types with the logic I implemented, I thought a single universal constructor was simpler for the caller to use. Also, since there are two constructors currently, the other supports opening a KML by resource ID, how could we implement opening a KMZ by resource ID? We'd have to imply the resource ID is either a KML or KMZ or we'd need the same logic to differentiate between them and call the appropriate stream constructor anyway.

library/src/main/java/com/google/maps/android/data/kml/KmlLayer.java
jeffdgr8: Yes, subdirectories are handled. The name includes the full file path within the zip archive.

library/src/main/java/com/google/maps/android/data/kml/KmlLayer.java
jeffdgr8: I added a similar method. It's not possible to pass a set of bitmap `InputStream`s, as each bitmap uses the same `ZipInputStream` as the `ZipEntry`s are iterated over. So I'm just passing a map of decoded bitmaps. https://github.com/googlemaps/android-maps-utils/pull/625/commits/482c748605b3d9a46a88118b1a9304d3bf77644d

library/src/main/java/com/google/maps/android/data/Renderer.java
jeffdgr8: Done https://github.com/googlemaps/android-maps-utils/pull/625/commits/482c748605b3d9a46a88118b1a9304d3bf77644d

library/src/main/java/com/google/maps/android/data/kml/KmlLayer.java
jeffdgr8: Added warning log: https://github.com/googlemaps/android-maps-utils/pull/625/commits/3d1d0e7c7f46ec08e3829ddb8a7e44e420ff6ff2

library/src/main/java/com/google/maps/android/data/kml/KmlLayer.java
arriolac: Good point about invoking by resource ID. Let's stick to this implementation.

PR #626, opened 2020-02-19
build(CodeCov): Enable CodeCov.

arriolac:
Enabling [CodeCov](https://codecov.io) on Travis so that code coverage reports can be generated and evaluated on a per-PR basis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,7 +96,7 @@ install:
   - source ~/.install-jdk-travis.sh
 
 script:
-  - if [ "$API" != "NONE" ]; then ./gradlew test check connectedCheck -x lint --stacktrace; fi
+  - if [ "$API" != "NONE" ]; then ./gradlew test check jacocoTestReport connectedCheck -x lint --stacktrace; fi
 
 before_cache:
   - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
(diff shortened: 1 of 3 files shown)


barbeau:
One of the Travis builds timed out, so I just restarted it (the rest passed).

.travis.yml
```diff
@@ -96,7 +96,7 @@ install:
   - source ~/.install-jdk-travis.sh
 
 script:
-  - if [ "$API" != "NONE" ]; then ./gradlew test check connectedCheck -x lint --stacktrace; fi
+  - if [ "$API" != "NONE" ]; then ./gradlew test check jacocoTestReport connectedCheck -x lint --stacktrace; fi
```
barbeau: `test` executes the JUnit tests we have in the `library/src/test`, and `connectedCheck` executes the instrumented tests we have in `library/src/androidTest`. Do you know if `jacocoTestReport` covers the output of both of these?

arriolac: `jacocoTestReport` will only evaluate the existence of a test (unit or instrumented test) for a particular set of classes (i.e. classes within the `:library` module). The actual result of the unit and instrumented tests are currently handled by Travis.

Example CodeCov result:
image

arriolac:
I misread the CodeCov report and thought that reports weren't being uploaded which is true for `master` since this change hasn't landed on `master` yet 😅 That said, no further changes are needed. cc @barbeau

PR #627, opened 2020-02-19
feat(cluster): Support updating existing items

### LINKED ISSUE #90: Update model and marker in cluster
When you put your model objects/items into the cluster manager, it will create the markers and group them to clusters if necessary. This is awesome!!

But what if the items change and upon this change you want to have different marker?
## Idea 1 (Ideal)

Update the api with an

```
 updateItem(Item)
```

method. Its the same instance of the added item but has updated attributes.
## Idea 2

What I first tried was just replacing the item list

```
getMap().clear();
mClusterManager.clearItems()
mClusterManager.addItems(items);
mClusterManager.cluster();
```

This worked only for visible clusters when map was zoomed out. However when map was zoomed in, visible map markers were removed by the above code. You have to zoom out until cluster appears and zoom in again to see the markers.
## Idea 3

Currently I am doing this, completely reinitiate the clusterManager when I need to update models and markers. I am not dealing with so much markers (max 200).

@broady Is this ok regarding performance/memory? Nevertheless the cluster api could need some rewrite :+1:

barbeau:
As discussed in https://github.com/googlemaps/android-maps-utils/issues/90, when an item instance is removed, updated, and re-added the `ClusterManager`, the updated item state wouldn't be reflected on the map.

The reason for this is that `DefaultClusterRenderer.CreateMarkerTask.perform()` was using a marker from its own `mMarkerCache` associated with that item to display that item on the map (even if `ClusterManager.clearItems()` was previously called). This resulted in the updated item contents not being reflected on the map, because the cached marker was not updated.

This PR does the following:
* Add new protected methods in `DefaultClusterRenderer` - `onClusterItemUpdated()` and `onClusterUpdated()` - which are called when an existing cached marker is found for an item/cluster when that item/cluster is being added back into the `ClusterManager`. The default implementation of these methods update the marker contents on the map, but then can also be overridden by applications for custom behavior (similar to `onBeforeClusterItemRendered()` and `onBeforeClusterRendered()`).  Following this change, you can now do `ClusterManager.clearItems()` and re-add the same list instance (or `remove()` and then `add()` the same item) and then call `cluster()` and the map markers will be updated with the new item model contents.
* Add a new `ClusterManager.updateItem()` helper method for updating the `ClusterManager` and algorithms with an existing item. The current implementation in `NonHierarchicalDistanceBasedAlgorithm` is a wrapper for `remove()` and then `add()`.
* Add more descriptive Javadocs to `ClusterManager`, `DefaultClusterRenderer`, and `NonHierarchicalDistanceBasedAlgorithm` to describe expected behavior, especially that `cluster()` should be invoked after any changes to `ClusterManager` to see changes on the map.

One new TODO I added that could potentially be addressed - it seems like the code to set the marker text when creating the item marker should be refactored into `onBeforeClusterItemRendered()` to match the implementation of clusters (`onBeforeClusterRendered()`), and to allow applications extending `DefaultClusterRenderer` to inherit this default behavior via a call to `super.onBeforeClusterItemRendered()`. However, moving this code would change the behavior of the library for apps already extending `DefaultClusterRenderer` that aren't calling `super.onBeforeClusterItemRendered()` - the marker title and snippet would no longer be updated. Thoughts on this are welcome!

Fixes #90

### DIFF
--- a/library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
+++ b/library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
@@ -16,13 +16,9 @@
 
 package com.google.maps.android.clustering;
 
-import android.content.Context;
-import android.os.AsyncTask;
-
 import com.google.android.gms.maps.GoogleMap;
 import com.google.android.gms.maps.model.CameraPosition;
 import com.google.android.gms.maps.model.Marker;
-import com.google.maps.android.collections.MarkerManager;
 import com.google.maps.android.clustering.algo.Algorithm;
 import com.google.maps.android.clustering.algo.NonHierarchicalDistanceBasedAlgorithm;
 import com.google.maps.android.clustering.algo.PreCachingAlgorithmDecorator;
@@ -30,6 +26,10 @@ import com.google.maps.android.clustering.algo.ScreenBasedAlgorithm;
 import com.google.maps.android.clustering.algo.ScreenBasedAlgorithmAdapter;
 import com.google.maps.android.clustering.view.ClusterRenderer;
 import com.google.maps.android.clustering.view.DefaultClusterRenderer;
+import com.google.maps.android.collections.MarkerManager;
+
+import android.content.Context;
+import android.os.AsyncTask;
 
 import java.util.Collection;
 import java.util.Set;
@@ -147,6 +147,10 @@ public class ClusterManager<T extends ClusterItem> implements
         return mAlgorithm;
     }
 
+    /**
+     * Removes all items from the cluster manager. After calling this method you must invoke
+     * cluster() for the map to be cleared.
+     */
     public void clearItems() {
         mAlgorithm.lock();
         try {
@@ -156,6 +160,11 @@ public class ClusterManager<T extends ClusterItem> implements
         }
     }
 
+    /**
+     * Adds items to clusters. After calling this method you must invoke cluster() for the state
+     * of the clusters to be updated on the map.
+     * @param items items to add to clusters
+     */
     public void addItems(Collection<T> items) {
         mAlgorithm.lock();
         try {
@@ -165,6 +174,11 @@ public class ClusterManager<T extends ClusterItem> implements
         }
     }
 
+    /**
+     * Adds an item to a cluster. After calling this method you must invoke cluster() for the state
+     * of the clusters to be updated on the map.
+     * @param myItem item to add to clusters
+     */
     public void addItem(T myItem) {
         mAlgorithm.lock();
         try {
@@ -174,6 +188,11 @@ public class ClusterManager<T extends ClusterItem> implements
         }
     }
 
+    /**
+     * Removes items from clusters. After calling this method you must invoke cluster() for the state
+     * of the clusters to be updated on the map.
+     * @param items items to remove from clusters
+     */
     public void removeItems(Collection<T> items) {
         mAlgorithm.lock();
         try {
@@ -183,6 +202,11 @@ public class ClusterManager<T extends ClusterItem> implements
         }
     }
 
+    /**
+     * Removes an item from clusters. After calling this method you must invoke cluster() for the state
+     * of the clusters to be updated on the map.
+     * @param item item to remove from clusters
+     */
     public void removeItem(T item) {
         mAlgorithm.lock();
         try {
@@ -193,7 +217,22 @@ public class ClusterManager<T extends ClusterItem> implements
     }
 
     /**
-     * Force a re-cluster. You may want to call this after adding new item(s).
+     * Updates an item in clusters. After calling this method you must invoke cluster() for the state
+     * of the clusters to be updated on the map.
+     * @param item item to update in clusters
+     */
+    public void updateItem(T item) {
+        mAlgorithm.lock();
+        try {
+            mAlgorithm.updateItem(item);
+        } finally {
+            mAlgorithm.unlock();
+        }
+    }
+
+    /**
+     * Force a re-cluster on the map. You should call this after adding, removing, updating,
+     * or clearing item(s).
      */
     public void cluster() {
         mClusterTaskLock.writeLock().lock();
--- a/library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
+++ b/library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
@@ -149,7 +149,7 @@ public class ClusterManager<T extends ClusterItem> implements
 
     /**
      * Removes all items from the cluster manager. After calling this method you must invoke
-     * cluster() for the map to be cleared.
+     * {@link #cluster()} for the map to be cleared.
      */
     public void clearItems() {
         mAlgorithm.lock();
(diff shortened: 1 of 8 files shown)


library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
```diff
@@ -156,6 +160,11 @@ public void clearItems() {
         }
     }
 
+    /**
+     * Adds items to clusters. After calling this method you must invoke cluster() for the state
```
arriolac: You can also make this more Javadoc friendly by linking to the `cluster()` method via `@link` (i.e. `"... you must invoke {@link #cluster()}"`). Same comment for other methods documented.

library/src/main/java/com/google/maps/android/clustering/algo/GridBasedAlgorithm.java
```diff
@@ -63,6 +63,12 @@ public void removeItems(Collection<T> items) {
         mItems.removeAll(items);
     }
 
+    @Override
+    public void updateItem(T item) {
+        removeItem(item);
```
arriolac: It would be good to check for the existence of `item` within `mItems` before invoking `removeItem` and `addItem`. Otherwise, there could be unintended effects. For example, if the developer did not not override `equals` and `hashCode` on the concrete `ClusterItem`, `item` will likely not be removed from the `Set` and a duplicate marker would be added.

Same comment for other concrete instances of `Algorithm`.

library/src/main/java/com/google/maps/android/clustering/view/DefaultClusterRenderer.java
```diff
@@ -737,34 +737,139 @@ public void remove(Marker m) {
...
     }
 
+    /**
+     * Called when a cached marker for a Cluster already exists on the map so the marker may
+     * be updated to the latest cluster values. Default implementation updated the icon with a
+     * circle with a rough count of the number of items. Note that the contents of the cluster may
+     * not have changed since the cached marker was created - implementations of this method are
+     * responsible for checking if something changed (if that matters to the implementation).
+     *
+     * The first time ClusterManager.cluster() is invoked on a set of items
```
arriolac: Nit: use Javadoc links `The first time {@link ClusterManager#cluster()}...`

library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
```diff
@@ -193,7 +217,22 @@ public void removeItem(T item) {
     }
 
     /**
-     * Force a re-cluster. You may want to call this after adding new item(s).
+     * Updates an item in clusters. After calling this method you must invoke cluster() for the state
+     * of the clusters to be updated on the map.
+     * @param item item to update in clusters
+     */
+    public void updateItem(T item) {
```
arriolac: Thoughts on returning a bool here? It could be helpful to know if an item you are trying to update was actually updated or not (and if not, can signal to the developer that they need to implement equals and hashcode).

library/src/main/java/com/google/maps/android/clustering/view/DefaultClusterRenderer.java
```diff
@@ -842,6 +947,7 @@ private void perform(MarkerModifier markerModifier) {
                         } else {
                             markerOptions.position(item.getPosition());
                         }
+                        // TODO (?) - Refactor set marker text into onBeforeClusterItemRendered()
```
barbeau: @arriolac Thoughts on refactoring the code below to `onBeforeClusterItemRendered()`? This would better match the implementation of `onBeforeClusterRendered()` for clusters and also allow applications extending `DefaultClusterRenderer` to inherit this default behavior along with a custom implementation via a call to `super.onBeforeClusterItemRendered()`. However, moving this code would change the behavior of the library for apps already extending `DefaultClusterRenderer` that aren't calling `super.onBeforeClusterItemRendered()` - the marker title and snippet would no longer be updated.

library/src/main/java/com/google/maps/android/clustering/view/DefaultClusterRenderer.java
barbeau: Done in https://github.com/googlemaps/android-maps-utils/pull/627/commits/556d20c2f9655569eaa2f972410e96ec6c381619

library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
barbeau: Done in https://github.com/googlemaps/android-maps-utils/pull/627/commits/556d20c2f9655569eaa2f972410e96ec6c381619.

library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
barbeau: :+1: That's a really good idea, and actually makes this testable too.  I went ahead and added returning boolean for all Algorithm add/remote/update operations to match Java Collection behavior in https://github.com/googlemaps/android-maps-utils/pull/627/commits/a1b1c17eac8d71ecbb0758bac6f1f514547c21f1, and added unit tests for them too. I could use another review on that just to make sure I didn't miss anything.

library/src/main/java/com/google/maps/android/clustering/algo/GridBasedAlgorithm.java
barbeau: :+1: I went ahead and added returning boolean for all Algorithm add/remote/update operations to match Java Collection behavior in https://github.com/googlemaps/android-maps-utils/pull/627/commits/a1b1c17eac8d71ecbb0758bac6f1f514547c21f1, and added unit tests for them too. I could use another review on that just to make sure I didn't miss anything.

barbeau:
Merging master branch to see if that enables CodeCov on this PR...

library/src/main/java/com/google/maps/android/clustering/view/DefaultClusterRenderer.java
arriolac: I think it only makes sense to refactor if the default implementation of updating title/snippet on the `MarkerOptions` is not desired. Seems that we'd always want that though. Can you think of a scenario where that might not be the case?

library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
```diff
@@ -156,44 +160,95 @@ public void clearItems() {
...
+     * state of the clusters to be updated on the map.
+     * @param items items to add to clusters
+     * @return true if the cluster manager contents changed as a result of the call
+     */
+    public boolean addItems(Collection<T> items) {
+        boolean result;
         mAlgorithm.lock();
         try {
-            mAlgorithm.addItems(items);
+            result = mAlgorithm.addItems(items);
```
arriolac: The `result` variable is not needed—the block within `finally` will still be invoked even after a `return`.

barbeau: Oh, good point - thanks for pointing this out. I just removed these for all the relevant methods and returned the result of `mAlgorithm.x()` directly in https://github.com/googlemaps/android-maps-utils/pull/627/commits/45af836aba6a1871a60150ee346c2bb183b298e6. Much cleaner.

library/src/main/java/com/google/maps/android/clustering/view/DefaultClusterRenderer.java
barbeau: I think this warrants it's own issue, especially given that the potential refactor would alter behavior - I'm going to remove the TODO in this PR and we can discuss in that thread. In short, yes - we actually override this behavior in the demo `CustomMarkerClusteringDemoActivity`.

PR #628, opened 2020-02-19
feat(ktx): GeoJson and Kml KTX

arriolac:
Took a pass in the `geojson` and `kml` packages to see what we can make Kotlin friendly. The only method I could find that we can improve is the constructor for `GeoJsonLayer` and `KmlLayer`.

_Before_:
```java
GeoJsonLayer layer = new GeoJsonLayer(map, geoJsonFile, null, polygonManager, null, groundOverlayManager);
```

_After_:
```kotlin
val layer = GeoJsonLayer(
    map = map,
    geoJsonFile = geoJsonFile,
    polygonManager = polygonManager,
    groundOverlayManager = groundOverlayManager
)
```

Other changes:
* making all extension functions and properties in the kx library `inline` to avoid overhead of new memory allocations when using the ktx aliases vs. using the method/properties directly
* Added `@ColorInt` annotations in `geojson` classes where I found it to be helpful
* Added `@RawRes` annotation in `kml` class
* Noticed an improvement that we can make in `KmlLayer`. Issue filed here #629

### DIFF
--- a/android-maps-utils-ktx/src/main/java/com/google/maps/ktx/kml/Kml.kt
+++ b/android-maps-utils-ktx/src/main/java/com/google/maps/ktx/kml/Kml.kt
@@ -1,12 +1,13 @@
 package com.google.maps.ktx.kml
 
+import android.content.Context
 import androidx.annotation.RawRes
-import androidx.fragment.app.FragmentActivity
 import com.google.android.gms.maps.GoogleMap
 import com.google.maps.android.collections.GroundOverlayManager
 import com.google.maps.android.collections.MarkerManager
 import com.google.maps.android.collections.PolygonManager
 import com.google.maps.android.collections.PolylineManager
+import com.google.maps.android.data.Renderer
 import com.google.maps.android.data.kml.KmlLayer
 import java.io.InputStream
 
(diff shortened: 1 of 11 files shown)


android-maps-utils-ktx/src/main/java/com/google/maps/ktx/kml/Kml.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.google.maps.android.data.kml.KmlLayer
+import java.io.InputStream
+
+/**
+ * Alias for the [KmlLayer] constructor that provides Kotlin named parameters and default values.
+ */
+inline fun kmlLayer(
+    map: GoogleMap,
+    @RawRes resourceId: Int,
+    activity: FragmentActivity,
```
barbeau: I'd suggest merging https://github.com/googlemaps/android-maps-utils/pull/631 and then update this to take in a `Context` and adding the cache parameter.

arriolac: Yep, agreed

arriolac: Updated fa1f839

PR #631, opened 2020-02-20
fix(KmlLayer): Update constructor to take in a Context instead

### LINKED ISSUE #629: Allow passing in a Context instead of a FragmentActivity when constructing KmlLayer
Currently, `KmlLayer` requires that a `FragmentActivity` is provided into its constructor before you are able to use this. The reason this is done is because the superclass of `KmlRenderer` (used internally in `KmlLayer`, `Renderer`, creates a retained Fragment so that images can be cached. 

Unfortunately, this is restrictive for the following reasons:
* it breaks clients that were previously used to passing in a `Context` instead
* while more than likely `KmlLayer` will be used in the context of a `FragmentActivity`, we cannot guarantee that clients have migrated to the `androidx` version of `FragmentActivity` (some may still be using `v4`)

A user should be able to pass in a `Context` instead of a `FragmentActivity` when constructing a `KmlLayer`. Image caching should be an optional optimization that should be injected in the `KmlLayer`.

Proposed `Renderer` signature:

```java
public Renderer(
    GoogleMap map, 
    Context context, 
    MarkerManager markerManager, 
    PolygonManager polygonManager, 
    PolylineManager polylineManager, 
    GroundOverlayManager groundOverlayManager,
    @Nullable ImagesCache imageCache
) {
    // ....
}
```

...this way, retaining `ImagesCache` is left to the caller and not implicitly handled.

 **Additional context**
Related PR #381

arriolac:
Updating `KmlLayer`, `KmlRenderer`, and `Renderer`, to accept a `Context` instead of a `FragmentActivity`. The former signature and solution presents a couple of issues mentioned in #629.

This change does break configuration change handling for caching images in a way that
retaining state across these changes is now handled by the caller (i.e. by passing in
an `ImagesCache` instance via `KmlLayer` constructor).

- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)

Fixes #629 🦕

### DIFF
--- a/demo/src/main/java/com/google/maps/android/utils/demo/MultiLayerDemoActivity.java
+++ b/demo/src/main/java/com/google/maps/android/utils/demo/MultiLayerDemoActivity.java
@@ -128,7 +128,7 @@ public class MultiLayerDemoActivity extends BaseDemoActivity {
         KmlLayer kmlPolygonLayer;
         try {
             // KML Polyline
-            kmlPolylineLayer = new KmlLayer(getMap(), R.raw.south_london_line_kml, this, markerManager, polygonManager, polylineManager, groundOverlayManager);
+            kmlPolylineLayer = new KmlLayer(getMap(), R.raw.south_london_line_kml, this, markerManager, polygonManager, polylineManager, groundOverlayManager, null);
             kmlPolylineLayer.addLayerToMap();
             kmlPolylineLayer.setOnFeatureClickListener(new KmlLayer.OnFeatureClickListener() {
                 @Override
@@ -140,7 +140,7 @@ public class MultiLayerDemoActivity extends BaseDemoActivity {
             });
 
             // KML Polygon
-            kmlPolygonLayer = new KmlLayer(getMap(), R.raw.south_london_square_kml, this, markerManager, polygonManager, polylineManager, groundOverlayManager);
+            kmlPolygonLayer = new KmlLayer(getMap(), R.raw.south_london_square_kml, this, markerManager, polygonManager, polylineManager, groundOverlayManager, null);
             kmlPolygonLayer.addLayerToMap();
             kmlPolygonLayer.setOnFeatureClickListener(new KmlLayer.OnFeatureClickListener() {
                 @Override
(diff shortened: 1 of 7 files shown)


demo/src/main/java/com/google/maps/android/utils/demo/MultiLayerDemoActivity.java
```diff
@@ -128,7 +128,7 @@ public void onFeatureClick(Feature feature) {
         KmlLayer kmlPolygonLayer;
         try {
             // KML Polyline
-            kmlPolylineLayer = new KmlLayer(getMap(), R.raw.south_london_line_kml, this, markerManager, polygonManager, polylineManager, groundOverlayManager);
+            kmlPolylineLayer = new KmlLayer(getMap(), R.raw.south_london_line_kml, this, markerManager, polygonManager, polylineManager, groundOverlayManager, null);
```
barbeau: Suggestion - in this or KmlDemoActivity, include an example of passing in the cache

library/src/main/java/com/google/maps/android/data/Renderer.java
```diff
@@ -127,29 +129,23 @@
      * Creates a new Renderer object for KML features
      *
      * @param map     map to place objects on
-     * @param activity activity needed to add info windows and retain bitmap cache fragment
+     * @param context the Context
      * @param markerManager marker manager to create marker collection from
      * @param polygonManager polygon manager to create polygon collection from
      * @param polylineManager polyline manager to create polyline collection from
      * @param groundOverlayManager ground overlay manager to create ground overlay collection from
```
barbeau: `imagesCache` should be added to Javadoc with a description

demo/src/main/java/com/google/maps/android/utils/demo/MultiLayerDemoActivity.java
arriolac: Ah, good point. I'll update `KmlDemoActivity` and reuse the `RetainFragment` + `ImagesCache` strategy there

PR #635, opened 2020-02-24
chore(Release): 1.0.0

arriolac:
Will tag once merged.


CHANGELOG.md
```diff
@@ -4,7 +4,32 @@ All notable changes to this project will be documented in this file. Dates are d
 
 Generated by [`auto-changelog`](https://github.com/CookPete/auto-changelog).
 
-#### [Unreleased](https://github.com/googlemaps/android-maps-utils/compare/0.6.2...HEAD)
+#### [Unreleased](https://github.com/googlemaps/android-maps-utils/compare/1.0.0...HEAD)
```
barbeau: I think we're missing a v1 header here? I believe this should be:

`#### [1.0.0](https://github.com/googlemaps/android-maps-utils/compare/0.6.2...1.0.0)`

...and then we should have another "Unreleased" section above it that would be:

`#### [Unreleased](https://github.com/googlemaps/android-maps-utils/compare/1.0.0...HEAD)`

CHANGELOG.md
```diff
@@ -4,7 +4,32 @@ All notable changes to this project will be documented in this file. Dates are d
...
+* **KmlRenderer:** Recurse over multi-geometry placemarks ([#605](https://github.com/googlemaps/android-maps-utils/issues/605)) ([8abcea8](https://github.com/googlemaps/android-maps-utils/commit/8abcea88f4039885dc87c94e63f3fd0b76069516))
+* increase stale bot window ([8811f97](https://github.com/googlemaps/android-maps-utils/commit/8811f97452db18c4262c4676c511fb2b0bb54961))
+
+### Features
+
+* **cluster:** Support updating existing items ([#627](https://github.com/googlemaps/android-maps-utils/issues/627)) ([39f0c22](https://github.com/googlemaps/android-maps-utils/commit/39f0c22e7c5619a92791024b1b31f7d9822f20a8)), closes [#90](https://github.com/googlemaps/android-maps-utils/issues/90) [#90](https://github.com/googlemaps/android-maps-utils/issues/90)
```
barbeau: #90 is linked twice here - given this is auto-generated, not sure if we want to change

arriolac: I'll remove since I make adjustments anyway after the changelog is autogenerated.

CHANGELOG.md
```diff
@@ -2,9 +2,34 @@
...
+> 24 February 2020
+
+### BREAKING CHANGES
+* Add MapObjectManager for multi-layer support with KML/GeoJSON renderers ([#380](https://github.com/googlemaps/android-maps-utils/pull/380))
+
+### Bug Fixes
+
+* **KmlLayer:** Update constructor to take in a Context instead ([#631](https://github.com/googlemaps/android-maps-utils/issues/631)) ([731f7e6](https://github.com/googlemaps/android-maps-utils/commit/731f7e61648ac1acd56770e0a841a172ade82847))
+* Maintain insertion order in clusters ([3ca965f](https://github.com/googlemaps/android-maps-utils/commit/3ca965f72c96fb44552c5410e6bdbc2815a82bac)), closes [#615](https://github.com/googlemaps/android-maps-utils/issues/615)
+* Make KML/GeoJSON lines and polygons clickable by default ([#614](https://github.com/googlemaps/android-maps-utils/issues/614)) ([122e96b](https://github.com/googlemaps/android-maps-utils/commit/122e96b8400af4c2f26f9c6237e3b4394b2a1d6a)), closes [#558](https://github.com/googlemaps/android-maps-utils/issues/558) [#454](https://github.com/googlemaps/android-maps-utils/issues/454) [#454](https://github.com/googlemaps/android-maps-utils/issues/454) [#558](https://github.com/googlemaps/android-maps-utils/issues/558)
```
barbeau: I just realized that #454 is linked twice here too

arriolac: ...and also #558. looks like the tool just finds all the linked issues and spits it out in the changelog (#558 and #454 are mentioned twice).

barbeau:
Also, https://github.com/googlemaps/android-maps-utils/pull/633 should technically be listed under "Breaking changes" because behavior changed.

arriolac:
> Also, #633 should technically be listed under "Breaking changes" because behavior changed.

Good catch—will update. Looks like the tool I used leaves out `refactor` type commit messages in the auto-generated changelog. A commit message with "BREAKING CHANGE" (number 3 in https://www.conventionalcommits.org/en/v1.0.0/) will catch though (left up to the person merging which is easy to forget).

> BREAKING CHANGE: a commit that has a footer BREAKING CHANGE:, or appends a ! after the type/scope, introduces a breaking API change (correlating with MAJOR in semantic versioning). A BREAKING CHANGE can be part of commits of any type.

barbeau:
@arriolac Ah, you're right - my bad for not including that in the commit message :man_facepalming:

PR #641, opened 2020-02-27
docs: Add snippet for docs

arriolac:


demo/build.gradle
```diff
@@ -39,15 +39,16 @@ android {
     }
 }
 
+// [START android_maps_utils_install_snippet]
 dependencies {
-    implementation project(':library')
```
barbeau: When building the demo app, this will always pull from Maven Central instead of building the library locally and using that dependency.  Is that the intent here?

demo/build.gradle
```diff
@@ -39,15 +39,16 @@ android {
     }
 }
 
+// [START android_maps_utils_install_snippet]
```
jpoehnelt:
```suggestion
// [START maps_android_utils_install_snippet]
```

region tags need to start with `maps_`

demo/build.gradle
arriolac: That's part of the intent. The other is so that this code snippet shows up in our documentation site so that whenever this version, this [page](https://developers.google.com/maps/documentation/android-sdk/utility/setup) is updated, too.

barbeau: Ok - could we leave `implementation project(':library')` in the file but commented out (basically inverting what was previously there)?

I know I frequently use the demo app to test changes to the library itself, and I'm guessing other contributors do too. Having the commented out text will help highlight that contributors shouldn't expect to see local library changes impact the demo app by default, and will also make it easy to manually toggle this when editing locally.

arriolac: Make sense. Will add back

arriolac: done in ffaf808

Issue #643, opened 2020-02-28
Update javadocs in http://googlemaps.github.io/android-maps-utils


arriolac:
**Is your feature request related to a problem? Please describe.**
Documentation in http://googlemaps.github.io/android-maps-utils/ is out-of-date.

 **Describe the solution you'd like**
(1) Page should be updated.
(2) Updating this page should be a part of the release process

 **Describe alternatives you've considered**
N/A

 **Additional context**
N/A

jpoehnelt:
also consider javadoc.io https://www.javadoc.io/doc/com.google.maps/google-maps-services/latest/index.html

arriolac:
> also consider javadoc.io https://www.javadoc.io/doc/com.google.maps/google-maps-services/latest/index.html

Nice. Once #644 is merged, the next published release should have docs available there (which should cover part (2) of this issue).

For (1), I've updated http://googlemaps.github.io/android-maps-utils/javadoc to display the most accurate version. Once a 1.x.x is released, we can reference https://www.javadoc.io/doc/com.google.maps.android/android-maps-utils in the README instead.

PR #644, opened 2020-02-28
docs: Add gradle task to generate docs.

arriolac:
A couple of things:
* adds task to generate docs i.e. `./gradlew javadoc`
* uploads `sources.jar` and `javadoc.jar` whenever we publish a new version of the lib

Updating http://googlemaps.github.io/android-maps-utils/javadoc/ is not yet automated. 

Relates to #643

### DIFF
--- a/gradle-mvn-push.gradle
+++ b/gradle-mvn-push.gradle
@@ -20,9 +20,20 @@ apply plugin: 'signing'
 archivesBaseName = 'android-maps-utils'
 group = 'com.google.maps.android'
 
+task javadoc(type: Javadoc) {
+    failOnError false
+    source = android.sourceSets.main.java.sourceFiles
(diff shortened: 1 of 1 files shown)


gradle-mvn-push.gradle
```diff
@@ -20,9 +20,20 @@ apply plugin: 'signing'
 archivesBaseName = 'android-maps-utils'
 group = 'com.google.maps.android'
 
+task javadoc(type: Javadoc) {
+    failOnError false
```
barbeau: I assume even though `failOnError false`, we'll still get some kind of error message if it fails saying it wasn't successful?

arriolac: Yep - errors and warnings will still be logged.

PR #667, opened 2020-03-19
docs: Update issue template

barbeau:
I've noticed that developers aren't providing some useful information in reports, like whether the issue they've seen occurs on a real device or emulator, and whether it can be reproduced using the demo app. I've updated the issue template to leave clear space for these details, and also removed some text that doesn't seem relevant to AMU.

I've also added clearer headings for expected vs. observed behavior.

@arriolac If any of this doesn't seem appropriate for this project, let me know and I can remove it.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -39,8 +39,8 @@ What did you see instead?  Describe your issue in detail here.
 #### Environment details
 
 1. Device make and model, or emulator details
-1. Android version
-1. Library version and other environment information
+2. Android version
+3. Library version and other environment information
 
 #### Steps to reproduce
 
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -24,13 +24,27 @@ Check for answers on StackOverflow with the [google-maps](http://stackoverflow.c
 
 Please be sure to include as much information as possible:
 
+#### Summary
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -24,14 +24,28 @@ Check for answers on StackOverflow with the [google-maps](http://stackoverflow.c
...
+What did you see instead?  Describe your issue in detail here.
+
 #### Environment details
 
-1. Specify the API at the beginning of the title (for example, "Places: ...")
-2. OS type and version
-3. Library version and other environment information
+1. Device make and model, or emulator details
+1. Android version
+1. Library version and other environment information
```
arriolac: Renumber as 1, 2, & 3?

barbeau: When this is rendered GitHub will convert these to actual numbers - it's the lazy persons way of numbering lists so you don't have to renumber then when items change 😉:

~~~
1. First
1. Second
1. Third
~~~

1. First
1. Second
1. Third

barbeau: Although the developer will see these unrendered 🤦‍♂. I'll change it now...

barbeau: Renumbered in https://github.com/googlemaps/android-maps-utils/pull/667/commits/d97174181d4db7f8f7673c51da3ff06a010c95e9

PR #669, opened 2020-03-19
fix: Fix including sources in release

### LINKED ISSUE #668: GeoJson: Cannot resolve in 1.0.1
Upon upgrading to version 1.0.1 of the library, I was no longer able to compile my code.  All references to classes within the com.google.maps.android.data.geojson package fail to resolve.

When I switch my version back to 1.0.0 or earlier.  Everything works fine.

#### Steps to reproduce

  1. Include the com.google.maps.android:android-maps-utils library version 1.0.1
  2. Attempt to compile code that references GeoJson classes

#### Code example

```
package com.example;

import android.app.Activity;
import android.os.Bundle;
import com.google.maps.android.data.geojson.GeoJsonPolygon;

public class MainActivity extends Activity
{

    @Override
    protected void onCreate(Bundle savedInstanceState)
    {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_main);
        GeoJsonPolygon polygon;
    }
}

```

#### Error
error: package com.google.maps.android.data.geojson does not exist
import com.google.maps.android.data.geojson.GeoJsonPolygon;

barbeau:
Reverts part of https://github.com/googlemaps/android-maps-utils/commit/7bbc3c7f9fb91e65264f6968f67dda17e78abc82.

Hopefully fixes #668 (?)

Also, going forward we should probably bump the demo app dependency immediately after publishing the library release artifact to potentially catch issues like this.

### DIFF
--- a/gradle-mvn-push.gradle
+++ b/gradle-mvn-push.gradle
@@ -29,6 +29,7 @@ task javadoc(type: Javadoc) {
 
 task sourcesJar(type: Jar) {
     from android.sourceSets.main.java.srcDirs
+    classifier = 'sources'
 }
 
 task javadocJar(type: Jar, dependsOn: javadoc) {


barbeau:
See https://github.com/googlemaps/android-maps-utils/issues/668#issuecomment-601384867:

>Its definentely that commit. I examined the pom's downloaded by gradle for the two versions, and 1.0.0 and below have a "<packaging>aar</packaging>" tag, whereas 1.0.1 has a "<packaging>pom</packaging>" tag. So its definently an issue in the maven push. Hopefully re-adding that line is enough to fix it.

arriolac:
Yikes... thanks for fixing. I _think_ this should do the trick as well. Let's push out `1.0.2` after this is merged.

barbeau:
@arriolac Could we merge https://github.com/googlemaps/android-maps-utils/pull/665 prior to `1.0.2` as well?

arriolac:
> @arriolac Could we merge #665 prior to `1.0.2` as well?

Yeah, let's merge that in, too.

RoarGronmo:
Confirming that 1.0.2 is ok.

barbeau:
@RoarGronmo Great, glad to hear it!

PR #672, opened 2020-03-19
chore: Bump demo app to library v1.0.2

barbeau:
Bump demo app to library v1.0.2


barbeau:
Seems like a good sign for 1.0.2 that this passed!

PR #674, opened 2020-03-20
docs: Update link to Javadoc

arriolac:
Now that we are uploading the `*-javadoc.jar` file, docs are automatically viewable at javadoc.io.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,10 @@ range of applications using the [Google Maps Android API][android-site].
 
 You can also find Kotlin extensions for this library [here][android-maps-ktx].
 
+## Developer Documentation
+
+You can view the generated [reference docs][javadoc] for a full list of classes and their methods.
+
 ## Requirements
 
 * Android API level 15+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,8 +52,7 @@ Or, if you'd like to contribute, send us a [pull request] and refer to our [code
 You can also reach us on our [Discord channel].
 
 For more information, check out the detailed guide on the
-[Google Developers site][devsite-guide]. You can also view the generated
-[reference docs][javadoc] for a full list of classes and their methods.
```
barbeau: I don't think `[javadoc]` was intended to be removed here?

arriolac: That was intentional since the text is in more prominent in its own section now. We can add that back if you think it'll help?

arriolac:
> A comment in-line.
> 
> Also, I don't see v1.0.2 docs at https://www.javadoc.io/doc/com.google.maps.android/android-maps-utils/latest/index.html? Would it still be propagating?

Hmm... odd. I had to visit this link first:

https://www.javadoc.io/doc/com.google.maps.android/android-maps-utils/1.0.2/index.html

and now it's visible at:

https://www.javadoc.io/doc/com.google.maps.android/android-maps-utils/latest/index.html

README.md
barbeau: Oh, sorry - somehow I missed that looking at the diff, LGTM.

PR #678, opened 2020-03-30
ci: Add GitHub Actions CI, reduce Travis matrix

barbeau:
~~*Please don't merge - work in progress*~~

Simple test with a single build configuration

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -1,0 +1,21 @@
+name: Android CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Build with Gradle
+      run: ./gradlew test check jacocoTestReport -x lint --stacktrace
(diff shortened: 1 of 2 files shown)


barbeau:
@arriolac A quick analysis is below - feedback is welcome.

### Performance

* [Travis](https://travis-ci.org/github/googlemaps/android-maps-utils/builds/668983268?utm_source=github_status&utm_medium=notification) - ~2m 30s each build (12 builds total)
* [GitHub Actions](https://github.com/googlemaps/android-maps-utils/actions/runs/66887558) - 3m 16s each build (1 build total)

However, the Travis build "ran for 9m 4s", but had "total time 30m 44s". So that means it was waiting for CPU time for over 20 minutes. We shouldn't have that level of wait time with GitHub Actions, to my knowledge.

### Reliability
The first GitHub Actions build did fail and I needed to restart it, and it succeeded on the 2nd build. I don't think this is typical, though - I haven't had any random failures for a desktop Gradle project that I've worked on for months, across dozens of builds. 

### Feature parity with Travis
It does appear that there is third-party support for running emulators on GitHub Actions:
* https://github.com/Malinskiy/action-android
* https://dev.to/ychescale9/running-android-emulators-on-ci-from-bitrise-io-to-github-actions-3j76

Also - without emulators do we really need a build matrix for different API levels, even on Travis? From a quick look at the `.travis.yml` none of the matrix variables are used, and, unless it's too late in the day and I should stop working, I can't think of why we would need a matrix without emulators. With Robolectric, we can always try to configure tests for a specific SDK - http://robolectric.org/configuring/.

arriolac:
@barbeau thanks for benchmarking. I agree that we don't need an extensive matrix if we are not even running tests on emulators. Also wouldn't hurt to have these run side-by-side and we can check in a few weeks from now if it makes sense to drop Travis.

barbeau:
@arriolac Ok, I've reduce the Travis matrix to a single build in https://github.com/googlemaps/android-maps-utils/pull/678/commits/bdaafc75d3daa0f9f58f2bc7c71728f8537d6ba6. Should be ready for review!

PR #691, opened 2020-04-13
docs: Adding javadoc for onClusterItemClick

arriolac:
Adding JavaDoc for onClusterItemClick. What returning true or false in this method does is a bit confusing so thought we'd document.

### DIFF
--- a/library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
+++ b/library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
@@ -363,6 +363,17 @@ public class ClusterManager<T extends ClusterItem> implements
      * Called when an individual ClusterItem is clicked.
      */
     public interface OnClusterItemClickListener<T extends ClusterItem> {
+
+        /**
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
```diff
@@ -363,6 +363,17 @@ public void setOnClusterItemInfoWindowClickListener(OnClusterItemInfoWindowClick
      * Called when an individual ClusterItem is clicked.
      */
     public interface OnClusterItemClickListener<T extends ClusterItem> {
+
+        /**
+         * Called when {@code item} is clicked.
+         *
+         * @param item the item clicked
+         *             
+         * @return true if the handler consumed the event (i.e. the default behavior should not
```
barbeau, line 372: If you want to match the `GoogleMap` docs exactly, `handler` would be `listener`. But, LGTM.

https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap.OnMarkerClickListener

PR #706, opened 2020-04-30
fix: Allow developers to use their own Maps API key in demo app

### LINKED ISSUE #705: Is an own API key strictly required now? Has anything changed?
Sorry, I don't have much experience in Android development, but so far, I was using an app (created just for my own, not published) based on this demo, and I was really happy to be able to do it without an API key.
Before that, if I just cloned the master branch and installed it to the phone, most of the features worked properly from the first moment, the map examples were visible and usable. But now the map is not loading. As far as I knew, Maps activities in native apps were allowed to use quite free. Has something changed on the side of Google? Is an API key required now, even if I just wanna show the map in the app? Thank you.

barbeau:
This PR changes the demo app from using a bundled Google Maps API key (which no longer seems to work - see #705) to allowing developers to provide their own Maps API key.

To run the demo app, you'll have to:

1. [Get a Maps API key](https://developers.google.com/maps/documentation/android-sdk/get-api-key)
1. Create a file in the `demo` directory called `secure.properties` (this file should *NOT* be under version control to protect your API key)
1. Add a single line to `demo/secure.properties` that looks like `MAPS_API_KEY=YOUR_API_KEY`, where `YOUR_API_KEY` is the API key you obtained in the first step
1. Build and run

TODO:
- [x] Add Toast if API key is empty

Fixes #705


barbeau:
I just realized I left out the Log statement if the API key is missing - I'm going to add that and then merge.

PR #721, opened 2020-05-14
feat: Add android-maps-utils-v3

### LINKED ISSUE #718: Support Maps SDK 3 beta
What are the plans for the Maps SDK 3 beta? That beta seems to be stagnant, but we can't implement it if the utils library doesn't support it.

arriolac:
Adding new module `library-v3` which is the module for the utilities library for Maps SDK for Android V3 BETA. This source code is autogenerated from code within the `library` module.

I also updated the publishing workflow so that it can be reused for both `library` and `library-v3` modules.

Fixes #718 🦕


README.md
```diff
@@ -38,7 +38,11 @@ You can view the generated [reference docs][javadoc] for a full list of classes
 
 ```groovy
 dependencies {
+    // Utilities for Maps SDK for Android
     implementation 'com.google.maps.android:android-maps-utils:1.2.1'
+
+    // Utilities for Maps SDK for Android V3 BETA
```
barbeau, line 44: I'd suggest some indication here that both dependencies shouldn't be used together - that may not be obvious to those unfamiliar with the SDK.

```suggestion
    // Alternately - Utilities for Maps SDK for Android V3 BETA
```

build.gradle
```diff
@@ -53,3 +56,174 @@ subprojects {
...
+
+    from('library/src/main/java') {
+        filter { line ->
+            line.replace('com.google.android.gms.maps', 'com.google.android.libraries.maps')
+        }
+        into 'main/java'
+    }
+
+    from('library/src/main/res') {
+        into 'main/res'
```
barbeau, line 76: Could there be any cases here where we need to do replacements of import packages in layouts? I can't think of any, but just wanted to call this out.

build.gradle
```diff
@@ -53,3 +56,174 @@ subprojects {
...
+    from('library/src/test') {
+        filter { line ->
+            line.replace('com.google.android.gms.maps', 'com.google.android.libraries.maps')
+        }
+        into 'test'
+    }
+    into 'library-v3/src'
+}
+
+///**
```
barbeau, line 88: Commented out code can be removed.

library-v3/src/main/res/values/styles.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
```
barbeau, line 1: If we're copying over everything in `/res`, then these files in `library-v3/src/main/res/` should be deleted.

build.gradle
```diff
@@ -53,3 +56,174 @@ subprojects {
...
+ */
+task copyGmsToV3Beta(type: Copy) {
+    group 'Custom'
+    description 'Generates source code for Maps SDK for Android V3 BETA Utilities'
+
+    from('library/src/main/java') {
+        filter { line ->
+            line.replace('com.google.android.gms.maps', 'com.google.android.libraries.maps')
+        }
+        into 'main/java'
```
barbeau, line 72: For each copy, should this be of the pattern `library-v3/src/main/java`?

Or does it pick up the module path elsewhere?

README.md
arriolac: Good idea. A similar edit should be made in the KTX lib.

build.gradle
arriolac: There weren't any gms specific imports in the layout file which is why I left out the package renaming filter. Think that should be added here (doesn't hurt)?

barbeau: I suppose it wouldn't hurt just in case things a future use case requires it. It might save someone in the future hunting down a bug due to mismatched package names.

arriolac: Makes sense. I'll add that here and add a comment why it was added.

Issue #741, opened 2020-06-01
Adding the dependency causes ClassNotFoundException


air89:
Hi all,

I was hoping to install the Maps Utils library to add dynamically labeled markers to my application. However, once I add the dependency to the gradle, my app no longer launches with the following issue:

```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.example.android.wealthplus, PID: 1122
    java.lang.RuntimeException: Unable to instantiate application com.example.android.wealthplus.StarterApplication: java.lang.ClassNotFoundException: Didn't find class "com.example.android.wealthplus.StarterApplication" on path: DexPathList[[zip file "/data/app/com.example.android.wealthplus-zNbUEd4nyAlGM6K9qnuwxQ==/base.apk"],nativeLibraryDirectories=[/data/app/com.example.android.wealthplus-zNbUEd4nyAlGM6K9qnuwxQ==/lib/arm64, /system/lib64, /product/lib64]]
        at android.app.LoadedApk.makeApplication(LoadedApk.java:1296)
        at android.app.ActivityThread.handleMakeApplication(ActivityThread.java:7007)
        at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6989)
        at android.app.ActivityThread.access$1600(ActivityThread.java:272)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2055)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:237)
        at android.app.ActivityThread.main(ActivityThread.java:8016)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1076)
     Caused by: java.lang.ClassNotFoundException: Didn't find class "com.example.android.wealthplus.StarterApplication" on path: DexPathList[[zip file "/data/app/com.example.android.wealthplus-zNbUEd4nyAlGM6K9qnuwxQ==/base.apk"],nativeLibraryDirectories=[/data/app/com.example.android.wealthplus-zNbUEd4nyAlGM6K9qnuwxQ==/lib/arm64, /system/lib64, /product/lib64]]
        at dalvik.system.BaseDexClassLoader.findClass(BaseDexClassLoader.java:196)
[...]

PR #743, opened 2020-06-05
chore: Custom InfoWindowAdapter demo and README.

### LINKED ISSUE #738: Custom Info Window stopped working in this new version. README does not explain anything
Before I start, I would like to say that I went through the other issues regarding this and how there is a new README with mitigation process. 

But the new README only explains if you use Marker Manager. 

How about the if you don't use marker manager at all?

this is how I add markers (I am using Glide)

```java
mUserMarker = mMap.addMarker(new MarkerOptions()
        .icon(BitmapDescriptorFactory.fromBitmap(bitmap))
         .position(loc)
         .title(getResources().getString(R.string.not_online))
```

and my custom info window

```java
mMap.setInfoWindowAdapter(new GoogleMap.InfoWindowAdapter() {
                @Override
                public View getInfoWindow(Marker marker) {
                    View myView;
                    myView = LayoutInflater.from(getActivity())
                            .inflate(R.layout.custom_user_info_window2,null, false);
                    return myView;
                }
                @Override
                public View getInfoContents(Marker marker) {
                    return null;
                }
            });
```

How to fix this in the new version where it stopped working.

Please help, thanks in advance

arriolac:
Add missing code to README on how to set a custom InfoWindowAdapter when using clustering.

Also added code in the demo project showing how this can be done.

Resolves #738

### DIFF
--- a/demo/src/main/java/com/google/maps/android/utils/demo/ClusteringDemoActivity.java
+++ b/demo/src/main/java/com/google/maps/android/utils/demo/ClusteringDemoActivity.java
@@ -16,10 +16,15 @@
 
 package com.google.maps.android.utils.demo;
 
+import android.view.LayoutInflater;
+import android.view.View;
+import android.widget.TextView;
 import android.widget.Toast;
 
 import com.google.android.gms.maps.CameraUpdateFactory;
+import com.google.android.gms.maps.GoogleMap;
 import com.google.android.gms.maps.model.LatLng;
+import com.google.android.gms.maps.model.Marker;
 import com.google.maps.android.clustering.ClusterManager;
 import com.google.maps.android.utils.demo.model.MyItem;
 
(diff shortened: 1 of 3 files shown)


demo/src/main/java/com/google/maps/android/utils/demo/ClusteringDemoActivity.java
```diff
@@ -43,6 +48,32 @@ protected void startDemo(boolean isRestore) {
...
+                textView.setText(text);
+                return view;
+            }
+
+            @Override
+            public View getInfoContents(Marker marker) {
+                return null;
+            }
+        });
+        mClusterManager.getMarkerCollection().setOnInfoWindowClickListener(new GoogleMap.OnInfoWindowClickListener() {
```
barbeau, line 69: Nit (kind of) - we could add Java 8 support to the demo app and use lambdas.

Generally, though, it may be a good idea to add Java 8 support to the app anyway to help developers avoid this issue - https://github.com/googlemaps/android-maps-utils/issues/741#issuecomment-638469115 - if they are copying some of the project properties, or just looking at the `build.gradle`.

arriolac: Good idea! Let me do that in a separate PR.

PR #752, opened 2020-06-18
feat!: Add support for setOnInfoWindowLongClickListener() for Markers

### LINKED ISSUE #750: Add support for setOnInfoWindowLongClickListener()
There is no setOnInfoWindowLongClickListener in the new utils.

barbeau:
As mentioned in #750, we currently support clicking on info windows for markers in the library, but not long clicking. This PR implements long clicking as well.

BREAKING CHANGE: Note that this could be considered a breaking change because `ClusterRenderer` is a `public interface` and may be used outside of the library if apps implemented their own renderer. Also, if apps were doing `map.setOnInfoWindowLongClickListener()` on their own this implementation will break that behavior and force them to use the collection info window long click listener, similar to how info window click listeners are currently handled (see the [Migration Guide](https://github.com/googlemaps/android-maps-utils#adding-a-custom-info-window)).

I've tested this in the demo app with:

~~~
        mClusterManager.getMarkerCollection().setOnInfoWindowLongClickListener(marker ->
                Toast.makeText(ClusteringDemoActivity.this,
                        "Info window long pressed.",
                        Toast.LENGTH_SHORT).show());
~~~

...and it works, although I can't add the above code in a PR until this is released and we bump the demo app version.

---

- [ ] Code coverage does not decrease (if any source code was changed)

Fixes #750 🦕

### DIFF
--- a/library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
+++ b/library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
@@ -16,6 +16,9 @@
 
 package com.google.maps.android.clustering;
 
+import android.content.Context;
+import android.os.AsyncTask;
+
 import com.google.android.gms.maps.GoogleMap;
 import com.google.android.gms.maps.model.CameraPosition;
 import com.google.android.gms.maps.model.Marker;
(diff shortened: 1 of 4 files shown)


bigbite95110:
All Connected........(NameTown or city That )

PR #754, opened 2020-06-22
docs: Add long click listener to cluster demo

barbeau:
Follow up to implementing the feature in https://github.com/googlemaps/android-maps-utils/pull/752.

### DIFF
--- a/demo/src/main/java/com/google/maps/android/utils/demo/ClusteringDemoActivity.java
+++ b/demo/src/main/java/com/google/maps/android/utils/demo/ClusteringDemoActivity.java
@@ -70,6 +70,10 @@ public class ClusteringDemoActivity extends BaseDemoActivity {
                 Toast.makeText(ClusteringDemoActivity.this,
                 "Info window clicked.",
                      Toast.LENGTH_SHORT).show());
+        mClusterManager.getMarkerCollection().setOnInfoWindowLongClickListener(marker ->
+                Toast.makeText(ClusteringDemoActivity.this,
+                        "Info window long pressed.",
+                        Toast.LENGTH_SHORT).show());
 
         try {
             readItems();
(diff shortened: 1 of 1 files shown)


demo/src/main/java/com/google/maps/android/utils/demo/ClusteringDemoActivity.java
```diff
@@ -70,6 +70,10 @@ public View getInfoContents(Marker marker) {
                 Toast.makeText(ClusteringDemoActivity.this,
                 "Info window clicked.",
                      Toast.LENGTH_SHORT).show());
+        mClusterManager.getMarkerCollection().setOnInfoWindowLongClickListener(marker ->
```
arriolac, line 73: Should this be preferred over setting listeners through `mClusterManager.setOnClusterItemInfoWindowLongClickListener`/`mClusterManager.setOnClusterInfoWindowLongClickListener`?

barbeau: Hmm, good question. To be honest I was following the existing pattern for setOnInfoWindowClickListener() - let me take a closer look.

kikoso: @barbeau , may I apply this change? I think setting the listener directly on the Cluster makes more sense - I can update this file to reflect that.

barbeau: To be honest it's been a while since I've looked at this, and I recall it not being straightforward but I don't recall the details. I'll trust your judgment on what you think is the best path forward. 😁

kikoso: I have added the listener directly to the Cluster. @wangela , if you have some time a review here would be appreciated.

wangela (approved):

PR #757, opened 2020-06-24
docs: Update Maps SDK v3 support in README

### LINKED ISSUE #756: Possibility clustering is integrated into maps sdk?
I noticed you mentioned

> Maps SDK via Google Play Services (this library is not yet compatible with the Maps SDK v3.0 BETA library)

It's a bummer that a majority of these features are not available within the new version of the maps sdk.  With apple maps introducing clustering in iOS 11, and applications wanting visualize large amounts of geospatial data has there been any talk about integrating things like clustering into the maps sdk.

barbeau:
* Call out more explicitly where Google Play Services is required
* Remove minor version from v3 (since we're now on v3.1 beta)

---

Closes #756

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,13 +32,13 @@ You can view the generated [reference docs][javadoc] for a full list of classes 
 ## Requirements
 
 * Android API level 15+
-* Maps SDK v2 via Google Play Services OR [Maps SDK v3 BETA] library
+* Maps SDK via Google Play Services OR [Maps SDK v3 BETA] library
 
 ## Installation
 
 ```groovy
 dependencies {
-    // Utilities for Maps SDK v2 for Android (requires Google Play Services) 
+    // Utilities for Maps SDK for Android (requires Google Play Services) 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,16 +32,16 @@ You can view the generated [reference docs][javadoc] for a full list of classes
 ## Requirements
 
 * Android API level 15+
-* Maps SDK via Google Play Services (this library is not yet compatible with the [Maps SDK v3.0 BETA] library)
+* Maps SDK v2 via Google Play Services OR [Maps SDK v3 BETA] library
```
arriolac, line 35: We're technically not branding this as "V2" (even though it is) in our documentation so I think we should keep this as "Maps SDK via Google Play Services" to avoid confusion.

barbeau: Sounds good - I just updated this in https://github.com/googlemaps/android-maps-utils/pull/757/commits/62a1cb24393cba434594bf05df37a8900634d352.

Issue #660, opened 2020-03-17
android-map-utils: Crash in ClusterTask


claudiofelber:
The ClusterTask doInBackground() method does sometimes crash. It seems that the write lock is released in a different thread than it is being acquired. The internal ReentrantReadWriteLock.tryRelease() method complains that the lock is not being held exclusively. This might have something to do with the fact that the ClusterTask is now running on the thread pool executor.

#### Environment details

Tested with android-map-utils 1.0.0 on Android API 28.

#### Steps to reproduce

Execute the cluster() method repeatedly with lots of items.

#### Stack trace
```
2020-03-17 13:56:31.860 29075-29148/ch.infoflora.florapp.dev E/AndroidRuntime: FATAL EXCEPTION: AsyncTask #1
    Process: ch.infoflora.florapp.dev, PID: 29075
    java.lang.RuntimeException: An error occurred while executing doInBackground()
        at android.os.AsyncTask$3.done(AsyncTask.java:354)
        at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:383)
        at java.util.concurrent.FutureTask.setException(FutureTask.java:252)
        at java.util.concurrent.FutureTask.run(FutureTask.java:271)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:764)
     Caused by: java.lang.IllegalMonitorStateException
        at java.util.concurrent.locks.ReentrantReadWriteLock$Sync.tryRelease(ReentrantReadWriteLock.java:373)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.release(AbstractQueuedSynchronizer.java:1285)
        at java.util.concurrent.locks.ReentrantReadWriteLock$WriteLock.unlock(ReentrantReadWriteLock.java:1138)
        at com.google.maps.android.clustering.algo.AbstractAlgorithm.unlock(AbstractAlgorithm.java:37)
        at com.google.maps.android.clustering.ClusterManager$ClusterTask.doInBackground(ClusterManager.java:297)
        at com.google.maps.android.clustering.ClusterManager$ClusterTask.doInBackground(ClusterManager.java:290)
        at android.os.AsyncTask$2.call(AsyncTask.java:333)
[...]

PR #761, opened 2020-07-02
fix: algorithm concurrency

jeffdgr8:
Lock and unlock a local algorithm reference. Ensure methods operate on the same algorithm reference during lock, operation, and unlock. Ensure algorithm reference is not replaced while in use.

This crash is caused by calling `ClusterManager.setAlgorithm()` immediately after `ClusterManager.cluster()` such that the `mAlgorithm` reference is changed in the middle of the background cluster task. So `mAlgorithm.lock()` and `mAlgorithm.unlock()` are actually referencing first the old and then the new algorithm. The thread that called `setAlgorithm()` is the one that locks the algorithm and the `AsyncTask` background thread is the one that attempts to unlock it, resulting in the crash.

This bug was introduced in https://github.com/googlemaps/android-maps-utils/pull/506 when the algorithm lock was moved from the `ClusterManager` to the `Algorithm`, which allows the algorithm to be accessed by a `ViewModel`, outside the `ClusterManager`, which can be recreated as part of the view layer.

Also restores https://github.com/googlemaps/android-maps-utils/pull/601 to use thread pools, since this wasn't the cause for this bug.

Fixes https://github.com/googlemaps/android-maps-utils/issues/660 🦕

### DIFF
--- a/library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
+++ b/library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
@@ -123,7 +123,7 @@ public class ClusterManager<T extends ClusterItem> implements
     public void setAlgorithm(ScreenBasedAlgorithm<T> algorithm) {
         algorithm.lock();
         try {
-            Algorithm<T> oldAlgorithm = getAlgorithm();
+            final Algorithm<T> oldAlgorithm = getAlgorithm();
             mAlgorithm = algorithm;
 
             if (oldAlgorithm != null) {
@@ -162,7 +162,7 @@ public class ClusterManager<T extends ClusterItem> implements
      * {@link #cluster()} for the map to be cleared.
      */
     public void clearItems() {
-        Algorithm<T> algorithm = getAlgorithm();
+        final Algorithm<T> algorithm = getAlgorithm();
         algorithm.lock();
         try {
             algorithm.clearItems();
@@ -178,7 +178,7 @@ public class ClusterManager<T extends ClusterItem> implements
      * @return true if the cluster manager contents changed as a result of the call
      */
     public boolean addItems(Collection<T> items) {
-        Algorithm<T> algorithm = getAlgorithm();
+        final Algorithm<T> algorithm = getAlgorithm();
         algorithm.lock();
         try {
             return algorithm.addItems(items);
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/google/maps/android/clustering/view/DefaultClusterRenderer.java
```diff
@@ -79,6 +81,7 @@
     private final ClusterManager<T> mClusterManager;
     private final float mDensity;
     private boolean mAnimate;
+    private final Executor mExecutor = Executors.newSingleThreadExecutor();
```
arriolac, line 84: Any reason this should be single threaded vs. a larger thread pool? Making this single threaded changes the current behavior/performance.

jeffdgr8: Yes, as noted by @barbeau https://github.com/googlemaps/android-maps-utils/pull/601#discussion_r369624658 when reviewing the original PR, the implementation is effectively single threaded already, which is why I chose to use a single thread executor. There's only ever one render task in flight and potentially one queued up to run next.

library/src/main/java/com/google/maps/android/clustering/ClusterManager.java
```diff
@@ -156,11 +162,12 @@ public void setAnimation(boolean animate) {
      * {@link #cluster()} for the map to be cleared.
      */
     public void clearItems() {
-        mAlgorithm.lock();
+        Algorithm<T> algorithm = getAlgorithm();
```
barbeau, line 165: Should these local declarations of `Algorithm` be declared `final`, to be safe?

jeffdgr8: Yeah, good call. That makes it more explicit they should remain constant through the entire method execution.

jeffdgr8: Done https://github.com/googlemaps/android-maps-utils/pull/761/commits/ee6d12d402591ba5daf53812dc576ae4c4c1aaee

library/src/main/java/com/google/maps/android/clustering/view/DefaultClusterRenderer.java
arriolac: Got it! That wasn't immediately obvious to me so this is definitely a nice improvement 👍

Issue #763, opened 2020-07-07
Fix publishing workflow


arriolac:
#### Summary

When `./gradlew publish` is executed, occasionally two separate uploads show up in the Sonatype dashboard resulting in missed artifacts (e.g. a `.aar` may sometimes be missing in a new version). This has manifested in developers not being able to download the version from maven. See #725. 

While this bug is still an issue, it's important for the person releasing to verify that all expected artifacts are part of a version release (i.e. `.aar`, `-sources.jar`, `-javadoc.jar`)

#### Expected behavior

Executing `./gradlew publish` once should upload `library` and `library-v3` module `.aar`s in a single releasable package on Sonatype.

#### Observed behavior

See Summary

#### Environment details

GitHub actions

#### Steps to reproduce

If you can reproduce using the demo app in the project, it helps us troubleshoot faster.

  1. Run `./gradlew publish` or execute the `publish.yml` GitHub workflow
  2. See Sonatype dashboard (only visible for repo owners) for multiple uploads (only sometimes reproducible)

PR #765, opened 2020-07-07
build: Use verbose over warn in publish workflow

arriolac:
Use `--verbose` over `--warn` to debug publish failures and address #763


arriolac:
> Whoops, looks like we crossed PRs in-flight - I opened #764 using `--debug`. I'm good keeping whichever you prefer.

`--debug` should work. Closing this.

PR #767, opened 2020-07-22
fix: Apply correction to distanceToLine() measurements

### LINKED ISSUE #720: PolyUtil.distanceToLine returns wrong result
### Summary
distanceToLine (from PolyUtil) returns wrong result, not the nearest point on the line will be found. 

here the test data:
**P**:  
longitude12.0978238, 
latitude: 49.3210674

**Start**:  
longitude: 12.097749 
latitude: 49.321045

**End**:
longitude: 12.097795
latitude: 49.321016

### Expected behavior
nearest distance from the point to the line (perpendicular). The distance should be ca. 5.5m
grafik

### Observed behavior
The green point is the calculated nearest point (su method variable), green is the line. Blue is the point. The distance is 5.956537

image

### Environment details
Java 

### Code example
 PolyUtil.distanceToLine(p, start, end). I checked this variable: su (method variable)

barbeau:
~**Please do not merge - work-in-progress**~

This is an implementation of a fix mentioned in https://github.com/googlemaps/android-maps-utils/issues/720#issuecomment-631263199 for a bug in the distance from a point to a line segment measurement.

TODO:
- [x] Implement correction
- [x] Additional unit tests to assess and document level of accuracy at different distances (see below)
    - Implement assertions of distance measurements between a point and a line segment at several orders of magnitude (e.g., 1, 10, 100, 1000, ... 1000000 meters) and at different latitude values (e.g., 0, 30, 60, 80) to assess the difference in distance accuracy as the distance and latitude change
    - All measurements used in unit tests should be between a point and a line *segment* (not the distance from a point to the line extending beyond the provided line end points) on the WGS84 ellipsoid (not straight lines on a projected flat surface). This can be calculated using tools like [QGIS](https://qgis.org/) and [ArcGIS](https://desktop.arcgis.com/), or using a library like GeoTools ([linestrings example](https://gis.stackexchange.com/questions/129375/using-linestring-distance-in-meters-with-geotools), [point to line example](https://gis.stackexchange.com/questions/132142/finding-distance-in-meters-between-point-and-path-in-lat-lon-using-geotools)) or Google's [S2geometry for Java](https://github.com/google/s2-geometry-library-java) (See https://s2geometry.io/about/overview for an example in C++ for measuring distance of a point to a polygon). A diagram showing the distance `D` that should be measured for each point is below.
    - Assertions should use the appropriate level of precision, and this precision should be documented in the Javadocs for the `distanceToLine()` method so developers understand the accuracy limitations of the current implementation. For example, if we only expect the algorithm to be accurate within 1 meter when measuring a distance of 10,000 meters between a line segment and point, there shouldn't be any decimal points in the `delta` value passed to `assertEquals()`. And the Javadocs should include a note about level of accuracy at this distance.

image

([Google Slide](https://docs.google.com/presentation/d/1Odus9d5y0ArUr9NZSnqMPpBs1UFm1Rv_Q51TO7kQc4E/edit?usp=sharing))

---

Fixes #720 🦕


alexaguilar89:
help me

alexaguilar89:
help

kikoso:
Hi @barbeau .

Can this PR be revived? I have been checking it and seems that solves the original problem.

What would it need, further test coverage?

barbeau:
Hey @kikoso! Yes, I'd love to see this get merged. See the unchecked checkbox in the original comment - I was hoping to add test coverage to confirm this fixed the issue and prevent it from getting broken again in the future. I don't currently have the bandwidth to work on this, but if you'd like to pick it up feel free.

kikoso:
Hi @barbeau .

I have added a few tests reflecting different latitudes and magnitudes. I have also added a case for a segment in a different hemisphere than the point, since this is an interesting edge case.

Regarding the accuracy: I can't find too much about how accurate the algorithm is, or what is the accuracy expected. Do we need to document this in the `distanceToLine()` function, since it is an algorithm specific issue? Do you have any pointers on where can I find more information about its accuracy?

Thanks.

barbeau:
>I have added a few tests reflecting different latitudes and magnitudes. I have also added a case for a segment in a different hemisphere than the point, since this is an interesting edge case. Distances have been verified using [QGIS](https://qgis.org/)

Thanks for this!! 💯 

>Regarding the accuracy: I can't find too much about how accurate the algorithm is, or what is the accuracy expected.

In the interest of getting this actually merged, I think as long as you've manually verified the measurements in QGIS that should be sufficient. IIRC the original thought was to include sufficient precision in the unit tests that someone could confirm it was accurate to X decimal places. To make sure future maintainers know this, can you please add a comment above those unit tests indicating they they were manually verified in QGIS?

Other than that, IMHO this should be good to merge now, although I'd welcome reviews by others.

kikoso:
@barbeau , that makes sense. I have added some comments providing a background introduction to this topic. Let me know what you think. Thanks!

barbeau:
LGTM, thanks!

kikoso (approved):

kikoso:
@wangela, feel free to take a last look.

wangela (approved):

PR #796, opened 2020-10-28
chore: Optimize computeLength() measurement

MxNxP:
The first step of the path iteration and the initialized value of the 'prev' point are the same, their distance is zero, ie it does not cause an error, but its calculation is unnecessary


MxNxP:
@googlebot I signed it!

barbeau (approved):
Thanks @MxNxP, LGTM!

PR #492, opened 2018-12-07
Made GeoJsonParser with parseGeometry public to use parser standalone

P72B:
As a user of the android-maps-utils lib I want to use the geoJson parser standalone to be able of parsing any kind of valid geoJson data with the lib.

PR #797, opened 2020-10-30
feat: Make GeoJsonParser fully public

### LINKED ISSUE #785: Make GeoJsonParser constructor public - allow FeatureCollection parsing
**Problem**
Unable to use `GeoJSONParser` to parse GeoJSON FeatureCollection

 **Solution**
The package private constructor of `GeoJSONParser` as seen below should be marked as public.
```java
    /**
     * Creates a new GeoJsonParser
     *
     * @param geoJsonFile GeoJSON file to parse
     */
    /* package */ GeoJsonParser(JSONObject geoJsonFile) {
        mGeoJsonFile = geoJsonFile;
        mGeoJsonFeatures = new ArrayList<>();
        mBoundingBox = null;
        parseGeoJson();
    }
```
 
**Alternatives considered**
Made an attempt to use `GeoJsonParser.parseGeometry()` but it ignores FeatureCollection geojsons and only seems to like pure geometry objects. 

 **Additional context**
* Previously requested here, but somehow missed (?): https://github.com/googlemaps/android-maps-utils/pull/492
* Please ensure GoogleMap (a UI thread-loving , non-serializable object) isn't required for data parsing classes such as Parsers.

barbeau:
In https://github.com/googlemaps/android-maps-utils/pull/492 the static `GeoJsonParser.parseGeometry()` method was made public based on requests from the community to be able to parse basic geometry objects.

However, as https://github.com/googlemaps/android-maps-utils/issues/785 points out this method ignores FeatureCollections and also doesn't handle styles, bounding boxes, and properties.

This patch makes the `GeoJsonParser` contructor and two getter methods (`getFeatures()` and `getBoundingBox()`) public so developers can use the `GeoJsonParser` in the same way that it's used internally within the library (instantiating it) without needing to instantiate a separate `GeoJsonLayer`.

Better Javadocs have also been added to `parseGeometry()` to explain how it can be used.

---

Closes #785 🦕

### DIFF
--- a/library/src/main/java/com/google/maps/android/data/geojson/GeoJsonParser.java
+++ b/library/src/main/java/com/google/maps/android/data/geojson/GeoJsonParser.java
@@ -105,7 +105,7 @@ public class GeoJsonParser {
      *
      * @param geoJsonFile GeoJSON file to parse
      */
-    /* package */ GeoJsonParser(JSONObject geoJsonFile) {
+    public GeoJsonParser(JSONObject geoJsonFile) {
         mGeoJsonFile = geoJsonFile;
         mGeoJsonFeatures = new ArrayList<>();
         mBoundingBox = null;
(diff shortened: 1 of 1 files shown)


arriolac (approved):
LGTM!

Issue #800, opened 2020-11-02
Build fails with Android Gradle Plugin 4.1.0


barbeau:
#### Summary

After updating from `classpath 'com.android.tools.build:gradle:4.0.2'` to `classpath 'com.android.tools.build:gradle:4.1.0'`, the demo app will no longer build and I get the error:

```
> Extension not initialized yet, couldn't access compileSdkVersion.
```

If I change back to `classpath 'com.android.tools.build:gradle:4.0.2'` it builds and runs without any problems.

Here's a branch showing the issue:

https://github.com/googlemaps/android-maps-utils/tree/sean/gradle-4.1.0

I believe this is an issue with Android Gradle Plugin 4.1.0, so I've opened an issue on the Google IssueTracker here:
https://issuetracker.google.com/issues/172273529

#### Expected behavior

Demo app should build (and from within Android Studio, the demo app should launch)

#### Observed behavior

```
Executing tasks: [:demo:assembleDebug] in project C:\android-projects\android-maps-utils

Starting Gradle Daemon...
Gradle Daemon started in 2 s 668 ms
Configuration on demand is an incubating feature.

FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':demo:compileDebugJavaWithJavac'.
> Extension not initialized yet, couldn't access compileSdkVersion.
```

####  Steps to reproduce

1. Clone/checkout https://github.com/googlemaps/android-maps-utils/tree/sean/gradle-4.1.0
1. Run the `demo` app from within Android Studio, or run `./gradlew :demo:assembleDebug` from the command line

#### Environment details

See branch https://github.com/googlemaps/android-maps-utils/tree/sean/gradle-4.1.0 for code.

Build: AI-201.8743.12.41.6858069, 202009231536, 

AI-201.8743.12.41.6858069, JRE 1.8.0_242-release-1644-b01x64 JetBrains s.r.o, OS Windows 10(amd64) v10.0 , screens 1536x960, 1707x960, 1536x864

AS: 4.1; Kotlin plugin: 1.4.10-release-Studio4.1-1; Android Gradle Plugin: 4.1.0; Gradle: 6.5; NDK: from local.properties: (not specified), latest from SDK: (not found); LLDB: pinned revision 3.1 not found, latest from SDK: (package not found); CMake: from local.properties: (not specified), latest from SDK: (not found), from PATH: (not found)

barbeau:
[...]

PR #818, opened 2021-01-27
chore: Use secrets_gradle_plugin.

### LINKED ISSUE #817: Update samples to use secrets-gradle-plugin
Not necessarily a problem but this FR results in a cleaner/simpler build.gradle file by pulling out common code and delegating it to a Gradle plugin.

Integrate [secrets-gradle-plugin](https://github.com/google/secrets-gradle-plugin) in the sample project for reading the GMP API key from a properties file. The plugin removes a lot of boilerplate code (i.e. reading from file, injecting values into BuildConfig, etc.)

Do nothing.

 **Additional context**
I've added the plugin in the android-samples project here https://github.com/googlemaps/android-samples/pull/432. I'm working on an additional feature that allows providing default keys to the plugin as a workaround for build failures related to keys missing.

arriolac:
Fixes #817 🦕


demo/build.gradle
```diff
@@ -16,9 +16,20 @@ import org.apache.tools.ant.filters.ConcatFilter
...
 
-apply plugin: 'com.android.application'
+plugins {
+    id 'com.android.application'
+    id 'com.google.secrets_gradle_plugin' version '0.4'
+}
 
 android {
+    signingConfigs {
+        tesing {
```
barbeau, line 26: Looks like this was left over from testing?

demo/v3/release/output-metadata.json
```diff
@@ -0,0 +1,20 @@
+{
```
barbeau, line 1: I don't think this or the .APK should be under version control?

local.defaults.properties
```diff
@@ -0,0 +1 @@
+MAPS_API_KEY=""
```
barbeau, line 1: Is this needed to avoid `null` values? I'm assuming so, but it's not clear if it's unused (I can't currently build the project).

build.gradle
```diff
@@ -23,7 +23,7 @@ buildscript {
         }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.0.2'
+        classpath 'com.android.tools.build:gradle:4.1.2'
```
barbeau, line 26: I still get https://github.com/googlemaps/android-maps-utils/issues/800 when trying to run the demo app of the gms variant on Android Studio 4.1.2 w/ plugin 4.1.2. Does this work on a different Android Studio version?

Does this require plugin 4.1.2, or could we revert back to plugin 4.0.2 for now?

barbeau (commented):
@arriolac This looks cool! Unfortunately I can't get it to run yet (see below), looks like due to the Gradle 4.1.2 plugin and #800.

demo/build.gradle
arriolac: Ah, yes. Will remove.

build.gradle
arriolac: The plugin does require 4.1.x of the Android Gradle Plugin and I'm also able to repro #800 using AS 4.1.2 although I could've sworn yesterday that I was able to get this to work. I also have the canary build installed so perhaps I was using that version of AS.

demo/v3/release/output-metadata.json
arriolac: Yep, agreed. Will remove.

local.defaults.properties
arriolac: This file is needed for builds to pass if the machine doesn't contain the MAPS_API_KEY in the secrets properties file. I explain the reasoning for a default properties file here: https://github.com/google/secrets-gradle-plugin/issues/6

build.gradle
arriolac: Looks like the same issue occurs on the canary build of AS. I was able to get the demo app to run using the `installGmsDebug` task though

arriolac:
> @arriolac This looks cool! Unfortunately I can't get it to run yet (see below), looks like due to the Gradle 4.1.2 plugin and #800.

Yeah, same :( Let's keep this open until #800 is addressed.

demo/build.gradle
```diff
@@ -16,7 +16,10 @@ import org.apache.tools.ant.filters.ConcatFilter
  * limitations under the License.
  */
 
-apply plugin: 'com.android.application'
+plugins {
+    id 'com.android.application'
+    id 'com.google.secrets_gradle_plugin' version '0.4'
```
barbeau, line 21: IIRC the coordinates and version number have changed since this initial PR?

demo/src/v3/java/com/google/maps/android/utils/demo/BaseDemoActivity.java
```diff
@@ -1,6 +1,6 @@
 /**
  * DO NOT EDIT THIS FILE.
- * <p>
+ *
```
barbeau, line 3: I believe if you rebase this it should get rid of a lot of these formatting changes in the v3 product flavor, given that https://github.com/googlemaps/android-maps-utils/pull/896 has been merged.

barbeau (commented):
@arriolac Now that https://github.com/googlemaps/android-maps-utils/pull/895 is merged this PR should be unblocked. A few comments in-line.

demo/build.gradle
arriolac: Yep—will update this PR with the new coordinates and version

arriolac:
@barbeau this is ready for your review when you get a chance. Note that I also updated the Android Gradle Plugin to the new recommended version—4.2.1.

barbeau (approved):
LGTM!

PR #828, opened 2021-03-22
Fix dependabot

hannesa2:
Maybe you observed it already, but dependabot is not working in this repository


.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "gradle" # See documentation for possible values
+    directory: "/" # Location of package manifests
```
arriolac, line 9: Should this be "/library"

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
...
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions"
```
arriolac, line 12: Nice!

arriolac (commented):
Thanks for the pull request. Just one question about the directory in your config. The previous config pointed to 2 modules (`library` and `demo`) so I want to confirm that the configuration you have set up will work for all modules.

.github/dependabot.yml
hannesa2: I would avoid this. Look at https://dependabot.com/java/

image

This pull request has more then current master, which doesn't work at all.

If you observe a missing scan, you can add it anytime.

arriolac (approved):

.github/dependabot.yml
hannesa2: Here is the evidence https://github.com/hannesa2/mlkit/pull/24/files

PR #831, opened 2021-03-23
chore: Combine release and publish workflow.

### LINKED ISSUE #829: Combine release and publish GitHub workflow
Currently the publish.yml workflow is not being executed from a tag pushed by the release.yml due changes preventing workflows from being started from another workflow. To fix this, the publishing workflow can be executed within the release.yml workflow such as done in https://github.com/googlemaps/android-maps-ktx/blob/main/.github/workflows/release.yml

arriolac:
The publishing workflow previously would start when a new Git tag was pushed on the repo and the release workflow was in charge of creating the tag.  However, due to some changes with GitHub actions, the previous set up no longer worked (i.e. another workflow that creates a Git tag cannot kick off another workflow). Consolidating releasing and publishing behavior fixes this issue, but is also a cleaner approach since publishing now occurs before a release is created on GitHub.

Fixes #829


barbeau (approved):
👍

PR #845, opened 2021-03-24
chore: Switching to main.

arriolac:
Building off of `main`. Once merged, I'll switch the default branch to `main` from `master`. More context: https://github.com/github/renaming

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -2,9 +2,9 @@ name: Android CI
 
 on:
   push:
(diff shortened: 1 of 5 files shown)


barbeau (commented):
👍  What's here LGTM, although I think there is another reference in `/.releaserc` that also needs to be changed:
https://github.com/googlemaps/android-maps-utils/blob/bb0fa692f83f5bd579cd4522d71ab8b3874b40e8/.releaserc#L2

Issue #858, opened 2021-04-16
Failing tests in release.yml


arriolac:
Release failed to run due to 102 failing tests. See: https://github.com/googlemaps/android-maps-utils/runs/2365872553?check_suite_focus=true

barbeau:
@arriolac It looks like the Gradle plugin got bumped to `4.1.3` by dependabot in https://github.com/googlemaps/android-maps-utils/commit/787006d5cb73439a347fc5ffc6d50cead99f8729 - could this be related to the same issue we saw in https://github.com/googlemaps/android-maps-utils/issues/800 when the build failed on `4.1.0`?

arriolac:
It could be related but looking at the logs, the error seems a bit different.

Not related to this issue but if the build passed due to dependabot bumping the Gradle plugin, should #800 be closed @barbeau ?

arriolac:
Looks like failures are due to Roboelectric:

```
2021-04-19T23:20:00.6610983Z         Caused by:
2021-04-19T23:20:00.6612264Z         org.apache.maven.artifact.resolver.MultipleArtifactsNotFoundException: Missing:
2021-04-19T23:20:00.6613982Z         ----------
2021-04-19T23:20:00.6614713Z         1) org.robolectric:android-all:jar:10-robolectric-5803371
2021-04-19T23:20:00.6615144Z 
2021-04-19T23:20:00.6615638Z           Try downloading the file manually from the project website.
2021-04-19T23:20:00.6616033Z 
2021-04-19T23:20:00.6616597Z           Then, install it using the command: 
2021-04-19T23:20:00.6617751Z               mvn install:install-file -DgroupId=org.robolectric -DartifactId=android-all -Dversion=10-robolectric-5803371 -Dpackaging=jar -Dfile=/path/to/file
2021-04-19T23:20:00.6618468Z 
2021-04-19T23:20:00.6618998Z           Alternatively, if you host your own repository you can deploy the file there: 
2021-04-19T23:20:00.6620395Z               mvn deploy:deploy-file -DgroupId=org.robolectric -DartifactId=android-all -Dversion=10-robolectric-5803371 -Dpackaging=jar -Dfile=/path/to/file -Durl=[url] -DrepositoryId=[id]
2021-04-19T23:20:00.6621208Z 
2021-04-19T23:20:00.6621571Z           Path to dependency: 
2021-04-19T23:20:00.6622223Z           	1) org.apache.maven:super-pom:pom:2.0
[...]

PR #860, opened 2021-04-20
chore: Add workflow_dispatch to all workflows.

arriolac:
In an attempt to debug issue #858, this change introduces two changes to the way GitHub workflows function in this repo:
* The `android` workflow will now run on all branches
* The `workflow_dispatch` was added so both `android` and `release` can be manually run. See: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -2,9 +2,10 @@ name: Android CI
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


barbeau (approved):
👍

PR #862, opened 2021-04-22
chore: Change android.yml to run ./gradlew build

arriolac:
Updating `android.yml` workflow to run the `build` task which will also run `check` and `test`. This change is related to debugging #858


barbeau (approved):
👍

PR #893, opened 2021-06-10
docs: Clarify NonHierarchicalViewBasedAlgorithm docs and fix demos

AndreasBoehm:
Clarify that width and height should be passed as dp to NonHierarchicalViewBasedAlgorithm

The calculations used to determine the visible bounds rely on the width and height being passed in as dp. The demo of the algorithm uses pixels which results in markers being rendered outside of the visible screen.

---

- [ ] Make sure to open a GitHub issue as a bug/feature request before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea

### DIFF
--- a/demo/src/gms/java/com/google/maps/android/utils/demo/ClusteringViewModelDemoActivity.java
+++ b/demo/src/gms/java/com/google/maps/android/utils/demo/ClusteringViewModelDemoActivity.java
@@ -53,7 +53,11 @@ public class ClusteringViewModelDemoActivity extends BaseDemoActivity {
 
         DisplayMetrics metrics = new DisplayMetrics();
         getWindowManager().getDefaultDisplay().getMetrics(metrics);
-        mViewModel.getAlgorithm().updateViewSize(metrics.widthPixels, metrics.heightPixels);
+
+        int widthDp = (int) (metrics.widthPixels / metrics.density);
+        int heightDp = (int) (metrics.heightPixels / metrics.density);
+
+        mViewModel.getAlgorithm().updateViewSize(widthDp, heightDp);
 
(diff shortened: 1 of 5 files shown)


AndreasBoehm:
@googlebot I signed it!

barbeau (changes requested):
@AndreasBoehm Good catch! Thanks for fixing this.

Two requests:
1. Could you also fix the variables passed to `updateViewSize()` in `ClusteringViewModelDemoActivity` to be dp instead of pixels?
2. Could you run the Gradle task `generateV3` to fix this issue in the `v3` demo app product flavor as well? This task will autogenerate the relevant `v3` code from the changes you already made to the `gms` variant here. Note that currently the `main` branch `v3` variant is already behind the `gms` variant with some whitespace and formatting changes - I've opened https://github.com/googlemaps/android-maps-utils/pull/896 to fix this.

AndreasBoehm:
> @AndreasBoehm Good catch! Thanks for fixing this.
> 
> Two requests:
> 
> 1. Could you also fix the variables passed to `updateViewSize()` in `ClusteringViewModelDemoActivity` to be dp instead of pixels?
> 2. Could you run the Gradle task `generateV3` to fix this issue in the `v3` demo app product flavor as well? This task will autogenerate the relevant `v3` code from the changes you already made to the `gms` variant here. Note that currently the `main` branch `v3` variant is already behind the `gms` variant with some whitespace and formatting changes - I've opened #896 to fix this.

Thanks for your review @barbeau.
Sorry it totally missed the usages of `updateViewSize()`. I pushed a new commit that fixes the variables and with updated `v3` code (excluded whitespace only changes)

barbeau (approved):
LGTM! Thanks @AndreasBoehm!

PR #898, opened 2021-06-24
chore: Support Kotlin as a development language.

### LINKED ISSUE #755: Add support for Kotlin in the library
This feature request is for the Maps SDK for Android Utility Library to support Kotlin in addition to Java. Kotlin has several modern language features (lambdas, nullability, coroutines, etc.) that make developing with it pleasing. Since increasingly more developers are using Kotlin, this is also a good way to continue catch a wider net of developers who may want to contribute to the library.

Add the following to the top-level `build.gradle` file:

```
// Project build.gradle file.
buildscript {
    ext.kotlin_version = '1.3.30'
    ...
    dependencies {
        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
    }
}
```

...and to each module:
```
// Inside each module using kotlin
apply plugin: 'kotlin-android-extensions'
apply plugin: 'kotlin-android'

...

dependencies {
   implementation "androidx.core:core-ktx:1.0.1"
   implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
}
```

N/A

 **Additional context**
Before considering integrating this, it would be good to assess the impact on the size of the `.aar` library by adding Kotlin support.

arriolac:
Add Kotlin dependency to modules so that Kotlin can be used in development alongside Java.

---

Fixes #755 🦕

### DIFF
--- a/demo/build.gradle
+++ b/demo/build.gradle
@@ -19,8 +19,8 @@ import org.apache.tools.ant.filters.ConcatFilter
 plugins {
     id 'com.android.application'
     id 'com.google.android.libraries.mapsplatform.secrets-gradle-plugin'
+    id 'kotlin-android'
 }
-apply plugin: 'kotlin-android'
 
 android {
     compileSdkVersion 30
--- a/demo/build.gradle
+++ b/demo/build.gradle
@@ -77,9 +77,8 @@ dependencies {
     v3Implementation 'com.android.volley:volley:1.2.0' // TODO - Remove this after Maps SDK v3 beta includes Volley versions on Maven Central
     v3Implementation 'com.google.maps.android:android-maps-utils-v3:2.2.4'
     v3Implementation 'androidx.multidex:multidex:2.0.1'
-    implementation "androidx.core:core-ktx:+"
     implementation "androidx.lifecycle:lifecycle-viewmodel-ktx:2.2.0"
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
     // [END_EXCLUDE]
 }
 // [END maps_android_utils_install_snippet]
@@ -125,7 +124,4 @@ secrets {
     // 2. Add this line, where YOUR_API_KEY is your API key:
     //        MAPS_API_KEY=YOUR_API_KEY
     defaultPropertiesFileName 'local.defaults.properties'
-}
-repositories {
-    mavenCentral()
 }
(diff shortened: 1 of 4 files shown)


library/build.gradle
```diff
@@ -61,6 +62,8 @@ dependencies {
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'org.robolectric:robolectric:4.5.1'
     testImplementation 'net.sf.kxml:kxml2:2.3.0'
+    implementation "androidx.core:core-ktx:+"
```
barbeau, line 65: I know use of `+` is generally discouraged because of lack of reproducibility between builds. Is there a reason it should be used here?

Same in other build.gradle file.

demo/build.gradle
```diff
@@ -122,3 +126,6 @@ secrets {
     //        MAPS_API_KEY=YOUR_API_KEY
     defaultPropertiesFileName 'local.defaults.properties'
 }
+repositories {
```
barbeau, line 129: We shouldn't need this here - it's declared in root `build.gradle`

demo/build.gradle
```diff
@@ -76,6 +77,9 @@ dependencies {
     v3Implementation 'com.android.volley:volley:1.2.0' // TODO - Remove this after Maps SDK v3 beta includes Volley versions on Maven Central
     v3Implementation 'com.google.maps.android:android-maps-utils-v3:2.2.4'
     v3Implementation 'androidx.multidex:multidex:2.0.1'
+    implementation "androidx.core:core-ktx:+"
+    implementation "androidx.lifecycle:lifecycle-viewmodel-ktx:2.2.0"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
```
barbeau, line 82: I think this should be `jdk8`? Same for other `build.gradle` files.

barbeau (commented):
Thanks @arriolac! A few comments in-line.

library/build.gradle
arriolac: Hmm weird. Looks like this is automatically added by Android Studio when you allow it to handle adding Kotlin dependencies for you. Same for the `repositories` block on each module.

barbeau:
@arriolac I believe we also need:

```
    kotlinOptions {
        jvmTarget = JavaVersion.VERSION_1_8
    }
```

...because we have:

```
    compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }
```

arriolac:
Ah good catch! @barbeau

barbeau (approved):
👍

Issue #917, opened 2021-07-08
Poor performance of StaticCluster#remove 


YegorKozlov:
NonHierarchicalDistanceBasedAlgorithm calls StaticCluster#remove  when re-clustering  candidates:

https://github.com/googlemaps/android-maps-utils/blob/862db8c3c3c3c4695bf3671b63ed8e05cce8657c/library/src/main/java/com/google/maps/android/clustering/algo/NonHierarchicalDistanceBasedAlgorithm.java#L212

StaticCluster internally stores the items in a ArrayList https://github.com/googlemaps/android-maps-utils/blob/862db8c3c3c3c4695bf3671b63ed8e05cce8657c/library/src/main/java/com/google/maps/android/clustering/algo/StaticCluster.java#L32 and the call of ArrayList#remove is O(N), i.e. every time we remove an item the list needs to iterate the elements until it finds the item.  This results in poor performance on large data sets (100K+ points)

A low-hanging fix is to  change the container from ArrayList to HashSet (or LinkedHashSet if ordering matters)

PR #918, opened 2021-07-08
perf: Use LinkedHashSet instead of List in StaticCluster

YegorKozlov:
Patch for https://github.com/googlemaps/android-maps-utils/issues/917


YegorKozlov:
@googlebot I signed it!

чт, 8 июл. 2021 г., 12:48 google-cla[bot] ***@***.***>:

barbeau (approved):
Thanks @YegorKozlov for pointing this out! And yes, I believe it does need to be a `LinkedHashSet`, because we already had an issue with the cluster contents shifting on re-cluster if insert order isn't maintained in other data structures (see https://github.com/googlemaps/android-maps-utils/pull/624).

PR #922, opened 2021-07-08
Gradle 6.9

hannesa2:
* 6.9 comes with native Apple Silicon support
* Verify Gradle action

PR #932, opened 2021-07-15
chore: Update Gradle wrapper to 6.9

barbeau:
This is the result of running:
```
./gradlew wrapper --gradle-version 6.9 --distribution-type all
```

This PR is a follow up to https://github.com/googlemaps/android-maps-utils/pull/922.

---


arriolac (approved):
:shipit:

PR #952, opened 2021-08-13
chore: Remove unused dex config

barbeau:
Build currently says:
>WARNING:DSL element 'dexOptions' is obsolete and should be removed.
 It will be removed in version 8.0 of the Android Gradle plugin.
 Using it has no effect, and the AndroidGradle plugin optimizes dexing automatically.

Additionaly, the "disablePreDex" property hasn't been used since we've switched to GitHub Actions. The link in the comments to the Android Tools site is also deprecated, so this practice no longer seems to be recommended.

---

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -46,24 +46,6 @@ task clean(type: Delete) {
 }
 
 /**
- * Improve build server performance by allowing disabling of pre-dexing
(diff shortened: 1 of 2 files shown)


arriolac (approved):
LGTM!

PR #999, opened 2021-10-04
chore: Remove android.yml

arriolac:
Removing `android.yml` workflow in favor of `test.yml` which is sync'd across other Android googlemaps libs.

Source of truth: https://github.com/googlemaps/.github/blob/master/sync-files/android/.github/workflows/test.yml


barbeau:
@arriolac Looks like the CI is hanging waiting on the `build` result?

>build Expected — Waiting for status to be reported

arriolac:
Should be fixed now. Had to update the branch protection rules to require `test.yml` to pass instead of `android.yml`.

barbeau (approved):
👍

Issue #772, opened 2020-08-05
Object order in MapObjectManager.Collection is random


solcott:
#### Summary

The order in which objects are stored in MapObjectManager.Collection and subclasses is random.  It is hard to compare the data stored here with other data sources for marker, polygon, etc.

#### Expected behavior

Storing data in a **LinkedHashSet** instead of a **HashSet** would keep the stored objects in the same order they were inserted

#### Observed Behavior

I have to create my own MapObjectManager, MarkerManager, etc to implement dragging Markers that also update a corresponding polygon.

#### Environment details

1. Device make and model, or emulator details - _Default Android 11 Emulator, Pixel 3, Samsung Tab 8_
2. Android version - _Android 8-11 at least, haven't tried other versions_
3. Library version and other environment information - _2.0.3 and previous versions back to 1.5_

jeffdgr8:
Fixed in #972

PR #1008, opened 2021-10-12
Keep object order in unmodifiable collection returned by MapObjectMan…

solcott:
Additional fix for #772 as it wasn't completely fixed by #972


kikoso:
Hello @solcott , @barbeau . What is the current status of this Pull Request? Should we still add some Unit Tests to cover the failing test? Does it make sense to get this merged, since the issue seems to be gone in the current `main` branch?

barbeau:
IIRC similar to my comment above, @jeffdgr8 made a good argument in https://github.com/googlemaps/android-maps-utils/pull/1008#discussion_r736085968 for still including this change due to it's impact on .equals() and .contains(), so we should add a unit test based on the code @jeffdgr8 included in https://github.com/googlemaps/android-maps-utils/pull/1008#discussion_r736085968 that currently fails on the main branch but passes after this change.

PR #1124, opened 2022-11-27
chore(dependencies): general update of dependencies

kikoso:
The current Pull Request introduces a general update of dependencies, and supersedes all the existing dependabot PRs.

This PR affecs the demo-java, demo-kotlin and snippets projects.

Increased compileSdkVersion.
Increased targetSdkVersion.
Moves namespace from AndroidManifest to the build.gradle file.
Increases Gradle plugin version.
General dependencies updated.
This supersedes all the existing dependabot PRs:

#1118, #1117, #1110, #1108 and #1069

### DIFF
--- a/demo/src/main/AndroidManifest.xml
+++ b/demo/src/main/AndroidManifest.xml
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
-    package="com.google.maps.android.utils.demo">
+    xmlns:tools="http://schemas.android.com/tools">
 
     <permission
         android:name="com.google.maps.android.utils.permission.MAPS_RECEIVE"
@@ -43,6 +42,7 @@
             android:value="${MAPS_API_KEY}" />
 
         <activity
+            android:exported="true"
             android:name=".MainActivity"
             android:label="@string/app_name">
             <intent-filter>
--- a/library/src/main/AndroidManifest.xml
+++ b/library/src/main/AndroidManifest.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.google.maps.android">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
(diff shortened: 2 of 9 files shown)


demo/src/main/AndroidManifest.xml
```diff
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
-    package="com.google.maps.android.utils.demo">
```
wangela, line 4: Would you comment why this attribute should be removed? Seems like a best practice changed to put the namespace in the build.gradle file?

library/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.google.maps.android">
```
wangela, line 3: Same here, curious why the package attribute should be dropped.

demo/src/main/AndroidManifest.xml
```diff
@@ -43,6 +42,7 @@
             android:value="${MAPS_API_KEY}" />
 
         <activity
+            android:exported="true"
```
wangela, line 45: Also curious about this change

demo/src/main/AndroidManifest.xml
kikoso: Hi @wangela. This has been deprecated in AGP 7.3.0, see official statement [here](https://developer.android.com/studio/releases/gradle-plugin#package-deprecated). The package should now be specified as a namespace in the Gradle file.

library/src/main/AndroidManifest.xml
kikoso: See comment [above](https://github.com/googlemaps/android-maps-utils/pull/1124#discussion_r1041903215).

demo/src/main/AndroidManifest.xml
kikoso: This PR targets Android 12+, and since then it is compulsory to specify the `android:exported` value. See official statement [here](https://developer.android.com/about/versions/12/behavior-changes-12#exported).

wangela (approved):

PR #1150, opened 2023-01-23
feat: added function to set text appearance for cluster

### LINKED ISSUE #1096: `DefaultClusterRenderer` - allow custom text appearance

I want to be able to customise the text appearance of the cluster marker, without having to implement/use my own `IconGenerator`. Similar to how I can customise the cluster color. Specifically, I want to be able to set the text color which sits on top of my color, in order to ensure I meet the required contrast for accessibility standards.

Expose a protected function in `DefaultClusterRenderer` which allows overriding the text appearance value. Something like:
```java
@StyleRes
protected int getClusterTextAppearance(int clusterSize) {
    return R.style.amu_ClusterIcon_TextAppearance; // Default value
}
```

This would allow the same level of customisation as the cluster color, by overriding this function and providing your own style.

_`getColor` included for reference:_
```java
protected int getColor(int clusterSize) {
    // ...
    return Color.HSVToColor(new float[]{
            hue, 1f, .6f
    });
}
```
 

Then setting the textAppearance of the built in `IconGenerator` as part of `getDescriptorForCluster`, e.g.:
```java
@NonNull
protected BitmapDescriptor getDescriptorForCluster(@NonNull Cluster<T> cluster) {
    int bucket = getBucket(cluster);
    BitmapDescriptor descriptor = mIcons.get(bucket);
    if (descriptor == null) {
        mColoredCircleBackground.getPaint().setColor(getColor(bucket));
        mIconGenerator.setTextAppearance(getClusterTextAppearance(bucket));
        descriptor = BitmapDescriptorFactory.fromBitmap(mIconGenerator.makeIcon(getClusterText(bucket)));
        mIcons.put(bucket, descriptor);
    }
    return descriptor;
}
```

Considerations with this approach:
- Does `setTextAppearance` have a performance overhead if setting for every `getDescriptorForCluster`? _As far as I can tell it doesn't, but I don't know if you have some benchmark tests or anything which could validate that?_

To customise the text appearance now, you have to:
- Create your own `IconGenerator`, configuring the content view
- Override `onBeforeClusterRendered` and `onClusterUpdated, replicating various code which has already been written within `DefaultClusterRenderer`

An example of this is already provided in `CustomMarkerClusteringDemoActivity`.

This does work for my use case, but it feels very heavy considering how easy it is to just override color.

 **Additional context**

While trying to work this out, I put a PR together ready which I am happy to submit if agreed.

**_Look forward to hearing some thoughts on this_**

kikoso:
The following PR introduces a function to set the text appearance for a cluster.

---

Fixes #1096 🦕

### DIFF
--- a/library/src/main/java/com/google/maps/android/clustering/view/DefaultClusterRenderer.java
+++ b/library/src/main/java/com/google/maps/android/clustering/view/DefaultClusterRenderer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013 Google Inc.
+ * Copyright 2023 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -38,6 +38,7 @@ import android.view.ViewGroup;
 import android.view.animation.DecelerateInterpolator;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.StyleRes;
 
 import com.google.android.gms.maps.GoogleMap;
 import com.google.android.gms.maps.Projection;
@@ -236,6 +237,12 @@ public class DefaultClusterRenderer<T extends ClusterItem> implements ClusterRen
         return Color.HSVToColor(new float[]{
                 hue, 1f, .6f
         });
+    }
+
+    @StyleRes
+    @Override
+    public int getClusterTextAppearance(int clusterSize) {
+        return R.style.amu_ClusterIcon_TextAppearance; // Default value
     }
 
     @NonNull protected String getClusterText(int bucket) {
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/google/maps/android/clustering/view/DefaultClusterRenderer.java
```diff
@@ -225,7 +226,13 @@ public int getColor(int clusterSize) {
         });
     }
 
-    @NonNull
```
wangela, line 228: May have been overwritten when resolving merge conflict. Should this be @NonNull too?

kikoso: Not, it is not really overwritten. This belongs to `getClusterText()`, which still keeps the `@NonNull` annotation. It is just a weird formatting from GitHub in the diff.

wangela (approved):

PR #1166, opened 2023-03-07
fix: Add container ground overlays in KML/KMZ

### LINKED ISSUE #740: GroundOverlay not put on map for KMZ-formatted bathymetric maps
I'm working on a drone app for boats, and have implemented the KmlOverlay to let users add bathymetric maps to their base maps. This works fantastic with the exception of KMZ files that include ground overlays. These are simply not displayed. I cross my fingers and hope you can fix this.

The app users typically utilize the "Reefmaster" software to create the bathymetric maps, and then export them to KMZ for use in other applications like Google Earth. 

In my app I simply add the layers to the map with these simple commands:
`layer = new KmlLayer(mMap, url, context, markerManager, polygonManager, polylineManager, groundOverlayManager, imageCache);
            layer.addLayerToMap();`

I found a workaround: Modifying the structure of the kml within the kmz (explained below). Then everything works. The modification is however messy: It alters and invalidates the root structure of the kml. So I would like not to advise app users to do this (it is a last resort for my beta users right now). Please note:
The original file imports perfectly into Google Earth, but KmlLayer fails to render the image. 
While the modyfied files fails to import in Google Earth, but now KmlLayer renders the image fine...

The root cause is related to how the KMZ from Reefmaster is structured. 

The KMZ file from Reefmaster typically contains two files:
- png image, if part of the map project
- doc.kml

The Reefmaster application typically exports to a KMZ with a KML containing 4 <Document> tags.
1. The styles
2. The majour contours
3. The minor contours
4. The shoreline

If users in Reefmaster add a so called "Shaded relied", a 3D-like png is added to the project below contours of the bathymetric map for improved readability. In the exported KML file, the groundoverlay part is added within the first Document tag (among the styles). Collapsed like this:

`kml ..
@xmlns:http://www.opengis.net/kml/2.2
Document ..
nameMyWater+
Style ..+
Style ..+
Style ..+
GroundOverlay ..+
Document ..
nameMajor Contours
+
Placemark ..+
Placemark ..+
+
Document ..
nameMinor Contours
+
Placemark ..+
Placemark ..+
+
Document ..
nameShorelines
+
Placemark ..`

The workaround I found is to cut out the ground overlay and put it above the first document tag. As explained, KmlLayer now works perfect. But the same file is not longer possible to import into Google Earth:
``kml ..
@xmlns:http://www.opengis.net/kml/2.2
GroundOverlay ..+
Document ..
nameMyWater+
Style ..+
Style ..+`

jeffdgr8:
Pass the correct set of ground overlays from a nested container to properly add to the map.

Fixes #740 🦕

### DIFF
--- a/library/src/main/java/com/google/maps/android/data/kml/KmlRenderer.java
+++ b/library/src/main/java/com/google/maps/android/data/kml/KmlRenderer.java
@@ -451,7 +451,7 @@ public class KmlRenderer extends Renderer {
             if (groundOverlayUrl != null && groundOverlay.getLatLngBox() != null) {
                 // Can't draw overlay if url and coordinates are missing
                 if (getCachedGroundOverlayImage(groundOverlayUrl) != null) {
-                    addGroundOverlayToMap(groundOverlayUrl, getGroundOverlayMap(), true);
+                    addGroundOverlayToMap(groundOverlayUrl, groundOverlays, true);
                 } else {
                     mGroundOverlayUrls.add(groundOverlayUrl);
                 }


OlavAa:
> Pass the correct set of ground overlays from a nested container to properly add to the map.
> 
> Fixes #740 🦕

Confirmed, issue is fixed.

wangela (approved):

wangela:
@jeffdgr8 Since this no longer uses `GetGroundOverlayMap()`, should this be treated as a breaking change?

jeffdgr8:
@wangela no, it shouldn't break anything. The current behavior uses `getGroundOverlayMap()` to get the root container's ground overlays and [passes them to `addGroundOverlays()`](https://github.com/googlemaps/android-maps-utils/blob/main/library/src/main/java/com/google/maps/android/data/kml/KmlRenderer.java#L129), along with the child containers. This function [iterates over the child containers](https://github.com/googlemaps/android-maps-utils/blob/main/library/src/main/java/com/google/maps/android/data/kml/KmlRenderer.java#L438) and adds their ground overlays as well. But the bug is that this function doesn't use these child containers' ground overlays from the parameter, but instead calls `getGroundOverlayMap()` to get the root container's ground overlays that have already been added.

So current behavior works for ground overlays in the root of the XML, but ground overlays in nested containers are not rendered as they should be. All ground overlays should be rendered correctly now with the change.

wangela:
Thank you @jeffdgr8 for contributing a fix to an issue reported by @OlavAa. Much appreciated!

PR #1168, opened 2023-03-29
docs: Uprev recommended android-maps-utils lib version

### LINKED ISSUE #1167: Docs reference wildly out of data android-maps-utils version
The first step in [the landing page docs](https://developers.google.com/maps/documentation/android-sdk/utility/setup) instruct the user to add the android-maps-utils dependency like so:
```
dependencies {
    implementation 'com.google.maps.android:android-maps-utils:2.3.0'

}
```
The current release version is `3.4.0`, though. The referenced version was broken in a few ways and caused some frustrating debugging until I spotted the error :)

aknobloch:
This fixes #1167. To be clear, I don't think this is a great fix. But I'm not familiar enough with the project to understand why the setup docs reference this `devsite` branch for pulling examples rather than either embedding it directly in the docs or using another method for dynamically updating things. So I've just elected to uprev the hardcoded value here.

The `devsite` branch does not currently build for many other reasons, and so:

1. I didn't try to make this branch build, nor run the tests.
2. I didn't try to manually search for other places in the docs or `devsite` branch which are out of date.

### DIFF
--- a/demo/build.gradle
+++ b/demo/build.gradle
@@ -61,7 +61,7 @@ android {
 
 // [START maps_android_utils_install_snippet]
 dependencies {
-    implementation 'com.google.maps.android:android-maps-utils:2.3.0'
+    implementation 'com.google.maps.android:android-maps-utils:3.4.0'
 
     // [START_EXCLUDE silent]
     implementation 'com.google.android.gms:play-services-maps:18.0.2'


bpr22715:
Hi @aknobloch ! Thanks for the PR.

We use semantic versioning for each PR. Could you make sure that the PR and commits are following the [SemVer structure](https://github.com/semantic-release/semantic-release)?

kikoso:
Hi @aknobloch! Thanks for the PR.

We use semantic versioning for each PR. Could you make sure that the PR and commits are following the [SemVer structure](https://github.com/semantic-release/semantic-release)?

aknobloch:
Thanks @kikoso - I'll work on making things conform to the semver guidelines.

>Also, we use region tags to update the website. You can probably change the target branch to googlemaps:main, and the tag will be updated.

This seems like a good idea, but looking at the [source](https://github.com/googlemaps/android-maps-utils/blob/main/demo/build.gradle) that particular tag doesn't have the project itself as a dependency, which makes sense.

Is there another way to reference the latest build version directly? I do think that this process of keeping the `devsite` branch up to date, especially for something like the core project version, is arduous.

aknobloch:
Friendly ping @kikoso.

kikoso:
Thanks @aknobloch! @wangela , are we planning to update the `devsite` soonish?

wangela (approved):

wangela:
Thank you for noticing and fixing @aknobloch! Hopefully I've also fixed our workflows to automate updating this branch in the future.

PR #1173, opened 2023-04-10
fix: Correctly parse `<hotSpot>` elements

### LINKED ISSUE #1172: Hotspot doesn't parse correctly
Styles containing a `<hotSpot>` tag are not parsed correctly. This is being caused by: https://github.com/googlemaps/android-maps-utils/blob/1ca8294288bc1d0752d13f55a24cc243737f7138/library/src/main/java/com/google/maps/android/data/kml/KmlStyleParser.java#L189-L191

A `<hotSpot>` element is always empty, it holds its data in attributes (see [the KML 2.3 schema](https://schemas.opengis.net/kml/2.3/ogckml23.xsd), look for `<element name="hotSpot"...`). This results in all `<hotSpot>` elements to not be parsed, and the anchor of the rendered Marker is the default anchor.

I have a pull request ready to go to fix it, I just saw in the PR template that I should create an issue before submitting.

SpecialMike:
Fixes #1172

The library currently never parses `<hotSpot>` element in KML, because they are always empty tags. A valid `<hotSpot>` element should be empty, with the necessary attributes like the following:
```xml
<hotSpot x="0.5" y="0.5" xunits="fraction" yunits="fraction"/>
```

I also adjusted the test for parsing `<hotSpot>` to be a valid KML file (none of `x`, `y`, `xunits`, or `yunits` are valid attributes for `<Style>`), and with a valid and invalid `<hotSpot`> in the case of an invalid `<hotSpot>` the default anchor should be used.

Finally, I also added some warnings to clarify that the library only supports `xunits` and `yunits` that have the value `"fraction"`. I would recommend that this clarification is made in the [KML support table in the official developer docs](https://developers.google.com/maps/documentation/android-sdk/utility/kml#supported). The warnings are the best way to let a user know about this, since there are other values that are part of the KML spec, and it's confusing why the `<hotSpot>` for a style isn't working when the units are `"pixels"` or `"insetPixels"`.

In order to test with another application, I had to update the root and library `build.gradle` files to support Jitpack. Let me know if I should take these changes out of the PR, but the current Jitpack override in the library `build.gradle` does not work.

### DIFF
--- a/library/src/main/java/com/google/maps/android/data/Style.java
+++ b/library/src/main/java/com/google/maps/android/data/Style.java
@@ -16,6 +16,8 @@
 
 package com.google.maps.android.data;
 
+import android.util.Log;
+
 import com.google.android.gms.maps.model.MarkerOptions;
 import com.google.android.gms.maps.model.PolygonOptions;
 import com.google.android.gms.maps.model.PolylineOptions;
@@ -30,6 +32,7 @@ import java.util.Observable;
  * and {@link com.google.maps.android.data.geojson.GeoJsonPolygonStyle GeoJsonPolygonStyle}
  */
 public abstract class Style extends Observable {
+    private static final String LOG_TAG = "Style";
 
     protected MarkerOptions mMarkerOptions;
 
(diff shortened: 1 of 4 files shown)


SpecialMike:
Sorry, I don't think I meant to include that change, it was just me trying to get this built to work with another project for testing. I force-pushed to remove that commit.

PR #1200, opened 2023-07-23
feat: added StreetViewUtil

### LINKED ISSUE #1201: Adding a utility library to check Street View availability
Right now, if we try to query a StreetMaps VIew location and it is not available, Google Map crashes (it returns a null value when null is not allowed).

A utility library could query the general availability of a location by calling the Street Maps View API, and handle it at a different level (returning an exception, a boolean, an enum with the availability of the location...)

See the following branch for a draft solution: https://github.com/googlemaps/android-maps-utils/tree/feat/added_street_view_util

Would you folks be ok with this utility supporting Coroutines? Would you prefer any other async mechanism?

kikoso:

Fixes #1201  🦕


wangela (changes requested):
The implementation LGTM! Please add usage documentation to the README to explain this utility's purpose and how to use it.

wangela (changes requested):
Please also add unit tests. Thank you!

wangela (approved):

PR #1235, opened 2023-10-09
fix: added value REQUEST_DENIED to StreetViewUtil

### LINKED ISSUE #1234: Add a REQUEST_DENIED value to the StreetMap Utility
StreetMaps can return a REQUEST_DENIED value when the Google Maps Key has not been properly set.

Adding this value to the StreetMaps Util would allow us to have a better error handling.

kikoso:
This PR adds the value REQUEST_DENIED to our StreetViewUtil, which allows us to handle this error. It happens when the authentication via API Key fails.

---

Fixes #1234 🦕

### DIFF
--- a/library/src/main/java/com/google/maps/android/StreetViewUtil.kt
+++ b/library/src/main/java/com/google/maps/android/StreetViewUtil.kt
@@ -82,5 +82,6 @@ data class ResponseStreetView(val status: Status)
 enum class Status {
     OK,
     ZERO_RESULTS,
-    NOT_FOUND
+    NOT_FOUND,
+    REQUEST_DENIED
 }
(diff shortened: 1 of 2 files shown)


wangela (approved):

library/src/main/java/com/google/maps/android/StreetViewUtil.kt
wangela, line 1: Let's add all possible status codes from documentation: https://developers.google.com/maps/documentation/streetview/metadata#status-codes

needs:
- OVER_QUERY_LIMIT
- INVALID_REQUEST
- UNKNOWN_ERROR

wangela (changes requested):

wangela (approved):

PR #1243, opened 2023-10-20
feat: added support for Advanced Markers

### LINKED ISSUE #1242: Add support for Advanced Markers
As a developer, I would like to have support for Advanced Markers in the Android Maps Utils

kikoso:
This PR updates our play-services-maps dependency to 18.2.0, introduces support for Advanced Markers and adds a sample.

---

- [ ] Will this cause breaking changes to existing Java or Kotlin integrations? If so, ensure the commit has a `BREAKING CHANGE` footer so when this change is integrated a major version update is triggered. See: https://www.conventionalcommits.org/en/v1.0.0/

Fixes #1242 🦕

### DIFF
--- a/demo/build.gradle
+++ b/demo/build.gradle
@@ -48,7 +48,7 @@ android {
 dependencies {
 
     // [START_EXCLUDE silent]
-    implementation 'com.google.android.gms:play-services-maps:18.1.0'
+    implementation 'com.google.android.gms:play-services-maps:18.2.0'
     implementation project(':library')
 
     implementation 'androidx.appcompat:appcompat:1.7.0-alpha01'
--- a/demo/build.gradle
+++ b/demo/build.gradle
@@ -48,7 +48,6 @@ android {
 dependencies {
 
     // [START_EXCLUDE silent]
-    implementation 'com.google.android.gms:play-services-maps:18.2.0'
     implementation project(':library')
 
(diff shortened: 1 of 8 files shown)


demo/build.gradle
```diff
@@ -48,7 +48,7 @@ android {
 dependencies {
 
     // [START_EXCLUDE silent]
-    implementation 'com.google.android.gms:play-services-maps:18.1.0'
+    implementation 'com.google.android.gms:play-services-maps:18.2.0'
```
wangela, line 51: We shouldn't need this line anymore because the library loads the Maps SDK dependency right?

library/src/main/java/com/google/maps/android/collections/MarkerManager.java
wangela, line 1: n.b. mixing `Marker`s with `AdvancedMarker`s should work, but is not recommended so we only demonstrate using one marker type at a time in the demo app.

No change needed, just noting this for future questions.

library/build.gradle
wangela, line 1: n.b. This converts the renderer to the new upgraded renderer by default per https://developers.google.com/maps/documentation/android-sdk/release-notes#October_18_2023

No change needed, just putting this here for release notes.

wangela (changes requested):

wangela (approved):

PR #1440, opened 2024-12-16
feat: make KML parser available as public API

nicolasygrand:
Allow library consumers to use `KMLParser` class for their project needs. Currently library consumers need to setup a functionnal view with GoogleMap object to validate KML data. Expose this class as public API solves this need.

### DIFF
--- a/library/src/main/java/com/google/maps/android/data/kml/KmlParser.java
+++ b/library/src/main/java/com/google/maps/android/data/kml/KmlParser.java
@@ -1,12 +1,12 @@
 /*
  * Copyright 2020 Google Inc.
- * 
+ *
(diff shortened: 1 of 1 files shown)


nicolasygrand:
> LGTM, this is a reasonable change.
> 
> @nicolasygrand , would you mind taking a look at the CLA? Did you sign it already?

Thanks for your feedback ;)

CLA is still at setup step in my company. Will take a look today.

nicolasygrand:
> Hi @nicolasygrand . This PR is a bit old. Did you manage to solve the CLA issue?

Unfortunately not. And this no longer a priority for my company. I will understand if you want to close this PR. :/

PR #1458, opened 2025-02-04
feat: setting animation on Cluster

kikoso:
This PR introduces a function to set the animation interpolator for the cluster. The current interpolator is lifted out of the Marker animation, since arguably we will have a single animation for the entire cluster. 

This change only applies to the `ClusterRendererMultipleItems` cluster, to gather some feedback before promoting it to the Abstract Cluster (which will introduce a breaking change)

### DIFF
--- a/library/src/main/java/com/google/maps/android/clustering/view/ClusterRendererMultipleItems.java
+++ b/library/src/main/java/com/google/maps/android/clustering/view/ClusterRendererMultipleItems.java
@@ -33,10 +33,15 @@ import android.os.Message;
 import android.os.MessageQueue;
 import android.util.SparseArray;
 import android.view.ViewGroup;
+import android.view.animation.AccelerateDecelerateInterpolator;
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/google/maps/android/clustering/view/ClusterRendererMultipleItems.java
```diff
@@ -85,10 +89,55 @@ public class ClusterRendererMultipleItems<T extends ClusterItem> implements Clus
...
+        EASE_OUT,
+        EASE_IN_OUT,
+        FAST_OUT_SLOW_IN,
+        BOUNCE,
+        ACCELERATE,
+        DECELERATE
+    }
+
+    void setAnimationType(AnimationType type) {
+        switch (type) {
```
dkhawk, line 109: Should this cancel any in progress animation?  Can that happen?

dkhawk (approved):

PR #1461, opened 2025-02-06
fix: fixed item updated

kikoso:
The following PR adds a line to update an individual item when the clustering has been re-triggered.


ullas-jain-eng (approved):

library/src/main/java/com/google/maps/android/clustering/view/ClusterRendererMultipleItems.java
```diff
@@ -31,6 +31,7 @@
 import android.os.Looper;
 import android.os.Message;
 import android.os.MessageQueue;
+import android.util.Log;
```
ullas-jain-eng, line 34: nit: This seems unused

ullas-jain-eng (approved):

dkhawk (approved):

PR #1559, opened 2025-08-13
feat: added ContinuousZoomEuclideanCentroidAlgorithm

kikoso:
This PR adds the `ContinuousZoomEuclideanCentroidAlgorithm `, a new clustering algorithm to fix some issues present in the `NonHierarchicalDistanceBasedAlgorithm` (namely, aggresivevity while clustering). It avoids the limitations of `NonHierarchicalDistanceBasedAlgorithm`, like abrupt clustering changes when zoom crosses integer levels and the square-based proximity check.

The algorithm works with continuous zoom, computing the clustering radius using the fractional zoom level, so the threshold changes smoothly as you zoom in or out. This makes clustering more natural and visually consistent.

It also uses Euclidean distance to decide cluster membership. The quadtree search still uses a square for performance, but items are only added to a cluster if they fall within a circular radius, avoiding the diagonal over-merging problem.

This comes as a separate algorithm instead of changing `NonHierarchicalDistanceBasedAlgorithm`, so we don’t introduce any behavior changes for existing users.


stefano-maggiolo (commented):
Looks good to me, but can we have a version that combine this and the Centroid version?

dkhawk (approved):

ColtonIdle:
Can we add some high level docs on when to use what? For example, it's pretty well discussed in this repo issues/prs that NonHierarchicalDistanceBasedAlgorithm is the goto when you have a ton of items, but maybe the readme can include a pros cons list or something? Thank you!

PR #1619, opened 2025-10-28
chore: Make DEFAULT_TOLERANCE public

dkhawk:
Make the `DEFAULT_TOLERANCE` constant in `PolyUtil` public, reverting an oversite when the class was ported to Kotlin.


kikoso (approved):
LGTM! This can be merged and released when we have some other meaningful PR.

PR #1677, opened 2026-04-16
fix: KMZ zip bomb mitigation by adding entry and size limits

kikoso:
Mitigates potential DoS attacks via KMZ 'zip bombs' by enforcing limits on:
- Total number of entries (max 200)
- Total uncompressed size (max 50MB)

A new test class  is added to verify these limits.

### DIFF
--- a/library/src/main/java/com/google/maps/android/data/kml/KmlLayer.java
+++ b/library/src/main/java/com/google/maps/android/data/kml/KmlLayer.java
@@ -46,6 +46,9 @@ import java.util.zip.ZipInputStream;
  */
 public class KmlLayer extends Layer {
 
+    private static final int MAX_KMZ_ENTRY_COUNT = 200;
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/google/maps/android/data/kml/KmlLayer.java
```diff
@@ -46,6 +46,9 @@
  */
 public class KmlLayer extends Layer {
 
+    private static final int MAX_KMZ_ENTRY_COUNT = 200;
```
dkhawk, line 49: Should these parameters be configurable with default settings?

dkhawk (approved):

library/src/main/java/com/google/maps/android/data/kml/KmlLayer.java
kikoso: This has been updated.

PR #1703, opened 2026-06-16
fix: enable filterTouchesWhenObscured on WebView

dkhawk:
Add the android:filterTouchesWhenObscured="true" attribute to the WebView tag in amu_webview.xml.


kikoso (approved):
LGTM!

PR #1710, opened 2026-06-24
fix: prevent StackOverflowError when parsing deeply nested KML containers and multi-geometries

dkhawk:
Add countdown max depth limits (default 20) when parsing nested KML containers (Folder, Document) and geometries (MultiGeometry) to prevent unbounded recursive descent from causing uncatchable StackOverflowErrors on deeply nested files.


kikoso (approved):
LGTM

LoyalAbbas (approved):

PR #1732, opened 2026-07-23
Redraw GeoJSON features after style changes

### LINKED ISSUE #1050: GeoJsonFeature.setPolygonStyle() doesn't redraw polygon if the feature was prewiously added via GeoJsonLayer.addFeature()
#### Summary

First I add GeoJsonLayer without features to map. Then I continiously add features using GeoJsonLayer.addFeature() and all the features are appearing on the map. Next I try to call GeoJsonFeature.setPolygonStyle() on the added features, but corresponding polygons don't change their appearance.

#### Expected behavior

Calling GeoJsonFeature.setPolygonStyle() should redraw the feature with the given style.

#### Observed behavior

The feature's appearance remains the same as before.

#### Environment details

Library version 2.3.0

#### Code example

```
// Add layer with empty features array
val geoJsons = JSONArray()
val geoJson = JSONObject()
geoJson.put("type", "FeatureCollection")
geoJson.put("features", geoJsons)
val layer = GeoJsonLayer(googleMap, geoJson)
val style = layer.defaultPolygonStyle
style.strokeColor = Color.RED
layer.addLayerToMap()

// This will draw the feature with red stroke
layer.addFeature(geoJsonFeature)

// Changing stroke color
val style = geoJsonFeature.polygonStyle
style.strokeColor = Color.BLUE
// This line should affect the polygon appearance but it doesn't
geoJsonFeature.polygonStyle = style
```

I suppose that the reason is Layer.addFeature() calls improper version of underlyed renderer's addFeature() (Renderer.addFeature() instead of GeoJsonRenderer.addFeature())

arpitagarwal1301:
Fixes #1050.

Observe legacy GeoJSON features so style and geometry changes invalidate the cached renderer model and redraw map objects with the updated values. Repeated insertion remains idempotent, removed features detach their observer, and null-geometry features remain safe across the layer lifecycle.

Multi-geometries now keep their rendered children under the parent feature, preserving complete removal and click lookup. Multi-polygons also carry their polygon style into the renderer.

Tests:
- data module unit tests, including focused redraw and renderer ownership regressions
- data module lint
- data debug assembly
- diff whitespace check

### DIFF
--- a/data/src/main/java/com/google/maps/android/data/geojson/GeoJsonLayer.kt
+++ b/data/src/main/java/com/google/maps/android/data/geojson/GeoJsonLayer.kt
@@ -26,14 +26,37 @@ import com.google.maps.android.collections.PolygonManager
 import com.google.maps.android.collections.PolylineManager
 import com.google.maps.android.data.Geometry
 import com.google.maps.android.data.Layer
+import com.google.maps.android.data.MultiGeometry
+import com.google.maps.android.data.parser.geojson.GeoJsonFeature as ParserGeoJsonFeature
+import com.google.maps.android.data.parser.geojson.GeoJsonFeatureCollection as ParserGeoJsonFeatureCollection
+import com.google.maps.android.data.parser.geojson.GeoJsonGeometry as ParserGeoJsonGeometry
+import com.google.maps.android.data.parser.geojson.GeoJsonGeometryCollection as ParserGeoJsonGeometryCollection
+import com.google.maps.android.data.parser.geojson.GeoJsonLineString as ParserGeoJsonLineString
+import com.google.maps.android.data.parser.geojson.GeoJsonMultiLineString as ParserGeoJsonMultiLineString
+import com.google.maps.android.data.parser.geojson.GeoJsonMultiPoint as ParserGeoJsonMultiPoint
+import com.google.maps.android.data.parser.geojson.GeoJsonMultiPolygon as ParserGeoJsonMultiPolygon
 import com.google.maps.android.data.parser.geojson.GeoJsonParser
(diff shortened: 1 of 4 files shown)


data/src/main/java/com/google/maps/android/data/geojson/GeoJsonLayer.kt
```diff
@@ -41,8 +42,11 @@ public class GeoJsonLayer : Layer {
     private var mRenderer: MapViewRenderer? = null
     private var mIsLayerOnMap = false
     private val mFeatureMap = HashMap<GeoJsonFeature, com.google.maps.android.data.renderer.model.Feature>()
-    private val mModelToLegacyFeatures = HashMap<com.google.maps.android.data.renderer.model.Feature, GeoJsonFeature>()
+    private val mModelToLegacyFeatures = java.util.IdentityHashMap<com.google.maps.android.data.renderer.model.Feature, GeoJsonFeature>()
```
kikoso: Can we use an import instead of the FQDN?

data/src/main/java/com/google/maps/android/data/geojson/GeoJsonLayer.kt
```diff
@@ -288,6 +291,18 @@ public class GeoJsonLayer : Layer {
                     )
                 }
 
+                is com.google.maps.android.data.renderer.model.MultiGeometry -> {
```
kikoso: Import instead.

kikoso (changes requested):

dkhawk:
@kikoso Thank you for the review. All inline fully qualified class names in GeoJsonLayer.kt have been replaced with imported aliases at the top of the file.

kikoso (approved):
LGTM!

PR #1743, opened 2026-08-05
build: make GPG signing conditional in PublishingConventionPlugin for local publishing

dkhawk:
### What this PR does

When developers run `./gradlew publishToMavenLocal` on their local machines without configuring GPG signing credentials (`signing.keyId`, `signing.secretKeyRingFile`), the build fails with:
```
Invalid publication 'maven': artifact file does not exist: '.../release-javadoc.jar.asc'
```
This occurred because `signAllPublications()` in `PublishingConventionPlugin.kt` was invoked unconditionally.

This PR wraps `signAllPublications()` so that it only signs when `signing.keyId` or `signing.secretKeyRingFile` is present and non-empty:
```kotlin
val isSigningConfigured = !providers.gradleProperty("signing.keyId").getOrElse("").isEmpty() ||
    !providers.gradleProperty("signing.secretKeyRingFile").getOrElse("").isEmpty()
if (isSigningConfigured) {
    signAllPublications()
}
```

### Verification
- Executed `./gradlew publishToMavenLocal` across all modules: **BUILD SUCCESSFUL** without `.asc` signature errors.
- Verified that all 7 library modules are correctly installed in `~/.m2/repository/com/google/maps/android/`.

### DIFF
--- a/build-logic/convention/src/main/kotlin/PublishingConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/PublishingConventionPlugin.kt
@@ -83,7 +83,19 @@ class PublishingConventionPlugin : Plugin<Project> {
             )
 
             publishToMavenCentral()
-            signAllPublications()
+
+            // Only enable GPG publication signing when signing credentials are present.
+            // Why: When signAllPublications() is invoked unconditionally, the Maven Publish plugin
+            // registers GPG signature files (.asc) as mandatory artifacts of the publication.
+            // On local developer machines without GPG keys configured, running tasks such as
+            // './gradlew publishToMavenLocal' fails because the .asc files cannot be generated.
+            // Making signing conditional allows local publishing to succeed without credentials
+            // while ensuring all CI and release pipeline builds are properly signed.
+            val isSigningConfigured = !providers.gradleProperty("signing.keyId").getOrElse("").isEmpty() ||
+                !providers.gradleProperty("signing.secretKeyRingFile").getOrElse("").isEmpty()
+            if (isSigningConfigured) {
+                signAllPublications()
+            }
 
             val artifactIdName = when (project.name) {
                 "maps-utils" -> "android-maps-utils"


kikoso:
Thanks for this @dkhawk, and sorry it sat open so long. The underlying problem was real, but it has since been fixed on `main` independently, so this PR is no longer needed.

`ca970251f` (#1766, the Kotlin migration PR, landed 2026-08-28) added an equivalent conditional at the same spot in `PublishingConventionPlugin.kt`:

```kotlin
publishToMavenCentral()
if (findProperty("signing.keyId")?.toString()?.isNotBlank() == true ||
    findProperty("signing.secretKeyRingFile")?.toString()?.isNotBlank() == true ||
    findProperty("signingInMemoryKey")?.toString()?.isNotBlank() == true
) {
    signAllPublications()
}
```

So `./gradlew publishToMavenLocal` without GPG credentials already succeeds on `main` today. That overlap is also why this PR now shows as conflicting.

Two reasons I'd keep the version on `main` rather than rebasing this one onto it:

1. It also checks `signingInMemoryKey`, vanniktech's in-memory signing path. Our `publish.yml` currently uses the keyring-file route (it seds `signing.keyId` and `signing.secretKeyRingFile` into `gradle.properties`), so this PR would not break the current release, but it would drop support for the in-memory path that `main` deliberately covers.
2. `isNotBlank` is a little safer than `!...isEmpty()` here, since `gradle.properties` ships these keys with empty values on lines 29-31 and CI rewrites them via `sed`. A stray space would flip an `isEmpty` check to true and reintroduce the original `.asc` failure.

One idea from here worth keeping though: `providers.gradleProperty(...)` is the configuration-cache-friendly Provider API, whereas `findProperty` reads project state at configuration time, which is exactly the pattern that trips up the configuration cache inside a convention plugin. If you want to land that separately, it is a small follow-up on `main`:

```kotlin
val signingProps = listOf("signing.keyId", "signing.secretKeyRingFile", "signingInMemoryKey")
if (signingProps.any { providers.gradleProperty(it).getOrElse("").isNotBlank() }) {
    signAllPublications()
}
```

Closing as already fixed.

PR #1758, opened 2026-08-22
fix(data): prevent SSRF from untrusted KML/GeoJSON icon and GroundOverlay href

herdiyana256:
## Summary

Icon and GroundOverlay `<href>` URLs are read from the parsed KML/GeoJSON document and fetched **without any validation**, giving a crafted (untrusted) document a **Server-Side Request Forgery (SSRF)** primitive.

- `Style/IconStyle/Icon/href` → `KmlStyle.iconUrl`
- `GroundOverlay/Icon/href` → `KmlGroundOverlay.imageUrl`

Both flow to `UrlIconProvider.loadBitmapFromUrl`, which did:

```kotlin
val url = URL(urlString)
val connection = url.openConnection() as HttpURLConnection
connection.doInput = true
connection.connect()          // <-- request issued; no validation anywhere
```

A malicious KML/GeoJSON (downloaded, shared, or user-provided) can therefore make the app issue requests to **loopback**, **link-local (`169.254.169.254` metadata)**, and **private-network** hosts reachable from the device. `HttpURLConnection` also **follows redirects by default**, so a public-looking href can `302` to an internal target.

The `KmlUrlSanitizer` interface exists but `sanitizeUrl(...)` is **never called anywhere in the codebase** (dead code), and `KmlLayer`/`GeoJsonLayer` hard-code `UrlIconProvider()` with no way to inject one — so there was **no protection by default and no way for a developer to add one**.

### Reproduction (network sequence is verbatim from `loadBitmapFromUrl`)

A KML with `<Icon><href>http://127.0.0.1:8474/INTERNAL</href></Icon>` (or a `169.254.169.254` / RFC1918 host) causes a real request to that host; a redirector href is followed to the internal target. Confirmed against a local listener that logged the inbound requests.

## Fix

- **`DefaultKmlUrlSanitizer`** (new): allows only `http`/`https` whose host does **not** resolve to a loopback / any-local / link-local / site-local (RFC 1918) / multicast address; returns `null` (block) otherwise. Public icon/overlay URLs are unaffected.
- **`UrlIconProvider`** now applies a `KmlUrlSanitizer` (secure default) before every fetch — blocked URLs issue no request — and sets `instanceFollowRedirects = false` to stop redirect-based bypass. Callers may pass a custom sanitizer, or `null` to opt out.
- **Hermetic unit tests** for the sanitizer (IP-literal hosts, no DNS/network).

This makes the previously-dead `KmlUrlSanitizer` functional and the default behavior secure, without changing the public `KmlLayer`/`GeoJsonLayer` API. Verified on a JVM: internal/metadata/RFC1918/redirector/`file://` hrefs are blocked with no request issued, while a public host still loads.

## Scope / notes

- DNS-rebinding between the sanitizer's resolution and the connection's resolution is a known residual limitation of host-based validation; disabling auto-redirects and failing closed on unresolved hosts substantially reduces the attack surface.

### DIFF
--- a/data/src/main/java/com/google/maps/android/data/kml/DefaultKmlUrlSanitizer.java
+++ b/data/src/main/java/com/google/maps/android/data/kml/DefaultKmlUrlSanitizer.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.maps.android.data.kml;
+
+import java.net.InetAddress;
+import java.net.URI;
+import java.util.Locale;
+
+/**
+ * Default, secure {@link KmlUrlSanitizer} used when a layer loads remote icons/overlays.
+ *
+ * <p>Icon and GroundOverlay {@code <href>} values come from the parsed KML/GeoJSON document,
+ * which is frequently untrusted (downloaded, user-provided, or shared). Fetching those URLs
+ * without validation is a Server-Side Request Forgery (SSRF) primitive: a crafted document can
+ * make the app issue requests to loopback, link-local (including the {@code 169.254.169.254}
+ * metadata endpoint), or private-network hosts reachable from the device.
+ *
+ * <p>This sanitizer allows only {@code http}/{@code https} URLs whose host does not resolve to a
+ * loopback, any-local, link-local, site-local (RFC 1918), or multicast address, and returns
+ * {@code null} (block) otherwise. Public icon/overlay URLs continue to load unchanged.
+ */
+public class DefaultKmlUrlSanitizer implements KmlUrlSanitizer {
+
+    @Override
+    public String sanitizeUrl(String url) {
+        if (url == null) {
+            return null;
+        }
+        final URI uri;
+        try {
+            uri = new URI(url);
+        } catch (Exception e) {
+            return null;
+        }
+
+        final String scheme = uri.getScheme();
+        if (scheme == null) {
+            return null;
+        }
+        final String lowerScheme = scheme.toLowerCase(Locale.ROOT);
+        if (!lowerScheme.equals("http") && !lowerScheme.equals("https")) {
+            return null;
+        }
+
+        final String host = uri.getHost();
+        if (host == null || host.isEmpty()) {
+            return null;
+        }
+
+        try {
+            // Block if ANY resolved address is internal, to defeat split-horizon / multi-A tricks.
+            for (InetAddress address : InetAddress.getAllByName(host)) {
+                if (address.isLoopbackAddress()
+                        || address.isAnyLocalAddress()
+                        || address.isLinkLocalAddress()
+                        || address.isSiteLocalAddress()
+                        || address.isMulticastAddress()) {
+                    return null;
+                }
+            }
+        } catch (Exception e) {
+            // Unresolvable host: fail closed.
+            return null;
+        }
+
+        return url;
+    }
+}
--- a/data/src/main/java/com/google/maps/android/data/kml/DefaultKmlUrlSanitizer.java
+++ b/data/src/main/java/com/google/maps/android/data/kml/DefaultKmlUrlSanitizer.java
@@ -28,9 +28,9 @@ import java.util.Locale;
  * make the app issue requests to loopback, link-local (including the {@code 169.254.169.254}
  * metadata endpoint), or private-network hosts reachable from the device.
  *
- * <p>This sanitizer allows only {@code http}/{@code https} URLs whose host does not resolve to a
- * loopback, any-local, link-local, site-local (RFC 1918), or multicast address, and returns
- * {@code null} (block) otherwise. Public icon/overlay URLs continue to load unchanged.
+ * <p>This sanitizer allows only {@code http}/{@code https} URLs whose host does not resolve to an
+ * internal address (see {@link #isInternalAddress(InetAddress)}), and returns {@code null} (block)
+ * otherwise. Public icon/overlay URLs continue to load unchanged.
  */
 public class DefaultKmlUrlSanitizer implements KmlUrlSanitizer {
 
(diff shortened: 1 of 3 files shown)


data/src/main/java/com/google/maps/android/data/kml/DefaultKmlUrlSanitizer.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        final String host = uri.getHost();
+        if (host == null || host.isEmpty()) {
+            return null;
+        }
+
+        try {
+            // Block if ANY resolved address is internal, to defeat split-horizon / multi-A tricks.
+            for (InetAddress address : InetAddress.getAllByName(host)) {
```
kikoso: This misses `fc00::/7` and `100.64.0.0/10/192.0.0.0/24`, so those internal ranges bypass the new sanitizer entirely. Would it be possible to cover them?

data/src/main/java/com/google/maps/android/data/renderer/UrlIconProvider.kt
```diff
@@ -35,6 +37,11 @@ import java.util.concurrent.ConcurrentHashMap
  */
 open class UrlIconProvider(
     private val dispatcher: CoroutineDispatcher = Dispatchers.IO,
+    // Icon/overlay hrefs originate from the (frequently untrusted) KML/GeoJSON document. The
+    // sanitizer is applied before every fetch to prevent SSRF; the secure default blocks
+    // non-http(s) schemes and hosts resolving to loopback/link-local/private ranges. Pass a
+    // custom implementation to widen/narrow the policy, or `null` to disable (not recommended).
+    private val urlSanitizer: KmlUrlSanitizer? = DefaultKmlUrlSanitizer(),
```
kikoso, line 44: This is a new parameter, can we use @JvmOverloads?

data/src/main/java/com/google/maps/android/data/kml/DefaultKmlUrlSanitizer.java
```diff
@@ -0,0 +1,81 @@
...
+                        || address.isMulticastAddress()) {
+                    return null;
+                }
+            }
+        } catch (Exception e) {
+            // Unresolvable host: fail closed.
+            return null;
+        }
+
+        return url;
```
kikoso, line 79: This validates the resolved `InetAddress` but returns the original hostname-based URL, not a pinned IP. `UrlIconProvider` then does `URL(safeUrl).openConnection()`, which re-resolves the hostname independently. An attacker-controlled DNS server can return a safe address to this check and a private/loopback address on the second lookup (classic DNS-rebinding, e.g. TTL=0), bypassing the sanitizer entirely regardless of how complete the range checks above are. Should the resolved address be pinned and reused for the actual connection, rather than re-resolving the hostname at fetch time?

data/src/main/java/com/google/maps/android/data/renderer/UrlIconProvider.kt
```diff
@@ -93,8 +100,18 @@ open class UrlIconProvider(
...
+                    if (urlSanitizer != null) {
+                        urlSanitizer.sanitizeUrl(urlString) ?: return@withContext null
+                    } else {
+                        urlString
+                    }
+                val url = URL(safeUrl)
                 val connection = url.openConnection() as HttpURLConnection
+                // Do not auto-follow redirects: a 30x to an internal host would otherwise bypass
+                // the sanitizer's host check (redirect-based SSRF).
+                connection.instanceFollowRedirects = false
```
kikoso: This silently breaks legitimate icon URLs that redirect (common for CDNs and signed S3/GCS URLs), the code never checks `connection.responseCode`, so a 3xx response just yields an empty/unexpected body, `BitmapFactory.decodeStream` returns `null`, and the icon fails with no exception or log. Since the goal is to stop redirect-based SSRF rather than to block redirects outright, should this instead follow one redirect manually and re-run it through `urlSanitizer.sanitizeUrl` before connecting?

kikoso:
Hi @herdiyana256 ! Thanks for the contribution, this fixes a real issue.

I have left a few comments, let me know if you want to discuss any of them.

herdiyana256:
Replied on each thread above. Short version: extra ranges (fc00::/7, 100.64.0.0/10, 192.0.0.0/24) covered, @JvmOverloads added, and redirects are now followed manually with a per-hop re-sanitize and a responseCode check. DNS-rebinding pin is the one open item, left a question about the approach on that thread. The failing check is the jacoco coverage-comment step missing a token on fork PRs, not a test failure.

data/src/main/java/com/google/maps/android/data/kml/DefaultKmlUrlSanitizer.java
herdiyana256: Covered. Moved the check into a static isInternalAddress and added byte checks for fc00::/7, 100.64.0.0/10 and 192.0.0.0/24 on top of the InetAddress ones, plus a fail-closed if getAllByName returns nothing. Added tests for each range.

data/src/main/java/com/google/maps/android/data/renderer/UrlIconProvider.kt
herdiyana256: Done, added @JvmOverloads to the constructor.

data/src/main/java/com/google/maps/android/data/kml/DefaultKmlUrlSanitizer.java
herdiyana256: Yeah, that's the real one. To pin I have to connect to the resolved IP but keep Host, SNI and cert verification on the original hostname, and SNIHostName is API 24+ while this module is minSdk 23, so it needs a guarded path. Want that here (custom SSLSocketFactory for SNI plus a hostname verifier bound to the original host, guarded for < 24), or as a follow-up? The range checks plus re-validating every redirect hop already close the direct and redirect vectors, so either works for me.

data/src/main/java/com/google/maps/android/data/renderer/UrlIconProvider.kt
herdiyana256: Changed it to follow up to 5 redirects by hand instead of blocking them. Each hop's Location goes back through sanitizeUrl before the next connect, and it now checks responseCode (2xx decode, 3xx follow, else bail), so a 3xx no longer decodes an empty body. Redirecting CDN/signed URLs load again.

data/src/main/java/com/google/maps/android/data/kml/DefaultKmlUrlSanitizer.java
kikoso: Let's not block the merge on it, it is currently good.

One correction on the API-level constraint though: you don't actually need `SNIHostName`/`SSLParameters.setServerNames()` (API 24+) to pin the IP while keeping the right SNI and cert verification. The older layered-socket approach works down to much lower API levels.

kikoso:
Thanks! Yeah, don't worry about the test flow, it doesn't run for remote contributors.

kikoso (approved):
LGTM

herdiyana256:
Thanks for the review and approval! Looks like the workflows are just waiting on a maintainer to approve the run for this fork PR. Whenever one of you has a moment to kick that off and merge, I'm happy to rebase if needed.

kikoso:
Ignore the reporting flow, doesn't work with external contributions

PR #1761, opened 2026-08-25
chore(clustering): use idiomatic poll().let in cluster renderers for null-safety

dkhawk:
### Summary

Replaces redundant `!queue.isEmpty()` checks and unsafe `.poll()` platform-type invocations in `performNextTask()` across cluster renderers with idiomatic Kotlin `queue.poll()?.let { ...; return }`.

### Details
- `Queue.poll()` is a Java platform method returning a nullable type (`AnimationTask?`). In strict nullability build environments (such as google3 with `-Xjsr305=strict`), calling `queue.poll().perform()` produces unsafe nullable receiver errors.
- In Kotlin, `queue.poll()` returns `null` if empty, making pre-checks with `!queue.isEmpty()` redundant. Using `poll()?.let` performs a single atomic queue retrieval, safely smart-casts `it` to non-null, and immediately returns once the highest-priority queued task is handled.

### Files Updated
- `DefaultAdvancedMarkersClusterRenderer.kt`
- `DefaultClusterRenderer.kt`
- `ClusterRendererMultipleItems.kt`

### Reviewer
cc @LoyalAbbas

### DIFF
--- a/clustering/src/main/java/com/google/maps/android/clustering/view/ClusterRendererMultipleItems.kt
+++ b/clustering/src/main/java/com/google/maps/android/clustering/view/ClusterRendererMultipleItems.kt
@@ -764,17 +764,11 @@ open class ClusterRendererMultipleItems<T : ClusterItem> @JvmOverloads construct
          * Perform the next task. Prioritise any on-screen work.
          */
         private fun performNextTask() {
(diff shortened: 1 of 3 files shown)


kikoso (approved):
LGTM

LoyalAbbas (approved):
LGTM

